PR #6, opened 2021-10-15
samples: point to the correct kata-deploy image

fidencio:
Instead of pointing to the latest released kata-deploy image, which does
not contain any work related to the confidential containers effort, let's
point to the `kata-deploy-cc:v0` one instead.

Signed-off-by: Fabiano Fidêncio <fabiano@fidencio.org>


bpradipt (approved):
/lgtm
Thanks @fidencio

PR #22, opened 2021-12-23
Add "CONFIGURE_CC=yes" env variable for kata-deploy image

bpradipt:
kata-deploy image uses the CONFIGURE_CC env variable to specifically add the CC config .


Jakob-Naucke (approved):
Thanks @bpradipt!

PR #31, opened 2022-03-30
Generalize install daemonset and extend by pre-install / post-uninstall hooks

jensfr:
With this PR I'm trying to remove Kata specifics from the controller code. Currently the code, especially the
part where we create the daemon set that runs the runtime installer, is very specific and tailored for kata-deploy.

What I'm trying to do is to factor out all runtime specific variables into custom resource fields and by that making
the code more generic.

On top of the refactoring this PR also brings a patch that extends the functionality of the operator. 

Patch 2 adds functionality to run code before the installation of the runtime and also right after it was uninstalled. This gives the opportunity to prepare the nodes and deploy any binaries, configuration files etc before the runtime is installed. The user can now specify a 

 - pre-install daemon set. Runs before the daemon set that install the container runtime
 - post-uninstall daemon set. Runs after the main daemon has finished uninstalling the container runtime

One example would be to replace the default containerd binary with one that has custom patches to run confidential containers [1]. This is done by specifying a daemonset in the CR yaml that does this. The config/samples/ccruntime.yaml file serves as an example and has all the definitions required to replace containerd, install the Kata containers runtime and eventually remove everything from the cluster. 

[1] a daemonset that replaces containerd with the patched one prepared by @fidencio


api/v1beta1/ccruntime_types.go
```diff
@@ -144,6 +144,85 @@ type CcInstallConfig struct {
...
+	// +optional
+	UninstallCmd []string `json:"uninstallCmd,omitempty"`
+
+	// This specifies the command for cleanup on the nodes
+	// +optional
+	CleanupCmd []string `json:"cleanupCmd,omitempty"`
+
+	// This specifies a node selector to define which nodes to install on
+	// +optional
+	CleanupNodeSelector *metav1.LabelSelector `json:"ccNodeSelector,omitempty"`
```
bpradipt, line 170: How about just using the CcNodeSelector field from CcRuntimeSpec to keep in sync ?

api/v1beta1/ccruntime_types.go
```diff
@@ -144,6 +144,85 @@ type CcInstallConfig struct {
...
+	// +optional
+	PreInstall PreInstallConfig `json:"preInstall,omitempty"`
+
+	// This specifies the configuration for the post-uninstall daemonset
+	// +optional
+	PostUninstall PostUninstallConfig `json:"postUninstall,omitempty"`
+}
+
+type PostUninstallConfig struct {
+	// This specifies the command executes before InstallCmd
```
bpradipt, line 186: nit: InstallCmd -> UninstallCmd

config/manager/kustomization.yaml
```diff
@@ -12,4 +12,5 @@ apiVersion: kustomize.config.k8s.io/v1beta1
 kind: Kustomization
 images:
 - name: controller
-  newName: quay.io/bpradipt/cc-operator
+  newName: quay.io/jensfr/confidential-containers-operator
```
bpradipt, line 15: How about changing it to generic values ?
```
  newName: controller
  newTag: latest
```

controllers/ccruntime_controller.go
```diff
@@ -88,7 +89,21 @@ func (r *CcRuntimeReconciler) Reconcile(ctx context.Context, req ctrl.Request) (
...
+		res, err := r.processCcRuntimeDeleteRequest()
+		if err != nil || res.Requeue {
+			r.Log.Info("error or requeue from processCcRuntimeDeleteRequest")
+			return res, err
+		}
+		if r.ccRuntime.Spec.Config.PostUninstall.Image != "" {
+			r.Log.Info("calling handlePostUninstall")
+			err, res := handlePostUninstall(r)
+			{
+				if err != nil && res.Requeue {
```
bpradipt, line 101: Should the conditional be 'or' - `||` ?

controllers/ccruntime_controller.go
```diff
@@ -142,6 +158,31 @@ func (r *CcRuntimeReconciler) processCcRuntimeDeleteRequest() (ctrl.Result, erro
...
+		return err, ctrl.Result{Requeue: true, RequeueAfter: time.Second * 10}
+	}
+
+	r.Log.Info("postuninstall image2", "r.ccRuntime.Spec.Config.Image",
+		r.ccRuntime.Spec.Config.PostUninstall.Image)
+
+	if r.ccRuntime.Spec.Config.PostUninstall.Image != "" &&
+		len(nodes.Items) < r.ccRuntime.Status.TotalNodesCount &&
+		r.ccRuntime.Status.TotalNodesCount > 0 {
+		r.Log.Info("inside != nil check")
```
bpradipt, line 174: Needed ?

controllers/ccruntime_controller.go
```diff
@@ -142,6 +158,31 @@ func (r *CcRuntimeReconciler) processCcRuntimeDeleteRequest() (ctrl.Result, erro
 	return ctrl.Result{}, nil
 }
 
+func handlePostUninstall(r *CcRuntimeReconciler) (error, ctrl.Result) {
```
bpradipt, line 161: Can you swap the return values to be in sync with other functions (ctrl.Result, error)? Makes it easier to read :-)

controllers/ccruntime_controller.go
```diff
@@ -181,6 +222,25 @@ func (r *CcRuntimeReconciler) processCcRuntimeInstallRequest() (ctrl.Result, err
...
+	err, nodes := r.getNodesWithLabels(map[string]string{"cc-preinstall/done": "true"})
+	if err != nil {
+		r.Log.Info("couldn't GET labelled nodes")
+		return ctrl.Result{}, err
+	}
+
+	if r.ccRuntime.Spec.Config.PreInstall.Image != "" &&
+		len(nodes.Items) < r.ccRuntime.Status.TotalNodesCount &&
+		r.ccRuntime.Status.TotalNodesCount > 0 {
+		r.Log.Info("inside != nil check")
```
bpradipt, line 235: Is the above log statement needed? or can it be reworded ?

controllers/ccruntime_controller.go
```diff
@@ -235,6 +295,39 @@ func (r *CcRuntimeReconciler) processCcRuntimeInstallRequest() (ctrl.Result, err
...
+	// if ds exists, get all labels
+	err, nodes := r.getNodesWithLabels(doneLabel)
+	if err != nil {
+		r.Log.Info("couldn't GET labelled nodes")
+		return ctrl.Result{Requeue: true, RequeueAfter: time.Second * 10}, err
+	}
+	if len(nodes.Items) < r.ccRuntime.Status.TotalNodesCount {
+		return ctrl.Result{Requeue: true, RequeueAfter: time.Second * 10}, err
+	}
+	err = r.Client.Delete(context.TODO(), foundPreinstallDs)
```
bpradipt, line 323: I'm wondering do we really need to delete the daemonsets. Any issues if we keep both the DS running to take care of node addition and removal to the cluster in a simpler way ?

config/manager/kustomization.yaml
jensfr: yes, of course, good catch, that shouldn't be in there. thanks!

controllers/ccruntime_controller.go
jensfr: yes, good catch

controllers/ccruntime_controller.go
jensfr: yes, makes sense, I'll change it

controllers/ccruntime_controller.go
jensfr: no, I'll remove it

controllers/ccruntime_controller.go
jensfr: no, it's a leftover, I'll remove it and look through the other log statements I added to clean everything up

controllers/ccruntime_controller.go
jensfr: That's a good point, I'll include scaling up and down nodes into my tests. My guess is that it will be better to only delete as part of the CCRuntime finalizer.

jensfr:
I pushed a new set of changes. Unfortunately the changes are quite big *and* I had to force push. Still, I tried to make it as easy to review as possible and also keep it bisectable. I recommend to review commit-by-commit.

jensfr:
by the way, to test it you need to use `image: quay.io/confidential-containers/container-engine-for-cc-payload:x86_64-2022042013051650452713` for preInstall.image and postUninstall.Image. The ":latest" tag doesn't work for me (any more). @fidencio

fidencio:
> by the way, to test it you need to use `image: quay.io/confidential-containers/container-engine-for-cc-payload:x86_64-2022042013051650452713` for preInstall.image and postUninstall.Image. The ":latest" tag doesn't work for me (any more). @fidencio

I think it's happening due to the image policy used, as `2022042013051650452713` is the same as `latest`, as shown here: https://quay.io/repository/confidential-containers/container-engine-for-cc-payload?tab=tags

config/samples/ccruntime.yaml
```diff
@@ -7,5 +7,118 @@ spec:
   # Add fields here
   runtimeName: kata
   config:
+    cleanupNodeSelector: 
+      - matchLabels:
+        - name: "katacontainers.io/kata-runtime"
+          value: "cleanup"
     installType: bundle
-    payloadImage: quay.io/confidential-containers/runtime-payload:v0
+    payloadImage: quay.io/kata-containers/kata-deploy:stable
```
fidencio, line 15: Just a note for our future selves that deserves documentation.  Right now this image does **not** have the CC bits in.
So, it'll work for our normal kata with the patched containerd, but it's not yet material for using the CC work that has been in development.

config/samples/ccruntime.yaml
```diff
@@ -7,5 +7,118 @@ spec:
...
+          path: /run/systemd
+          type: ""
+        name: systemd
+      - hostPath:
+          path: /usr/local/bin/
+          type: ""
+        name: local-bin
+    installCmd: ["/opt/kata-artifacts/scripts/kata-deploy.sh", "install"]
+    uninstallCmd: ["/opt/kata-artifacts/scripts/kata-deploy.sh", "cleanup"]
+    cleanupCmd: ["/opt/kata-artifacts/scripts/kata-deploy.sh", "reset"]
```
fidencio, line 60: Just a note, I see you've group Install / uninstall / cleanup altogether, and I like it.  I like it very much that I wonder if there's some way to do the same with preInstall / postUninstall.

This doesn't have to be done right now, but that's something we can do before having a release out.
So, postUninstall / preInstall would become something like `Dependency`, and we could group in `installDependencyCmd`, and `uninstallDependencyCmd`, which would simplify our CRD.

@jensfr, @bpradipt, WDYT?

fidencio (approved):
I've left two comments, but overall looks good to me.

Those comments can be addressed as a follow-up patch, and I can submit the follow-up patch if @bpradipt and @jensfr agree with the simplification suggested.

config/samples/ccruntime.yaml
jensfr: @fidencio it's a good idea. I need to think about if it has other implications but I'll give it a show. As you said I think in a follow-on pull request. Thank you for the review!

controllers/ccruntime_controller.go
jensfr: btw: I changed this now. The daemonsets stay running and are only deleted when the custom ressource is deleted altogether.

controllers/ccruntime_controller.go
jensfr: I changed that now

controllers/ccruntime_controller.go
jensfr: this one is also removed now

config/manager/kustomization.yaml
jensfr: I changed this also

api/v1beta1/ccruntime_types.go
jensfr: I removed the cleanupNodeSelector now, it's not used any more. However I just noticed that it is still in the example CR yaml. I'll remove it there as well and force push.

config/samples/ccruntime.yaml
bpradipt: Addressing it in a follow-on PR is a good approach. May be creating an issue for this will be a good reminder.

Issue #34, opened 2022-05-05
bitnami images are no longer available from quay.io


bpradipt:
Need to switch to dockerhub.
The following thread provides more detail - https://github.com/bitnami/bitnami-docker-postgresql/issues/336

PR #35, opened 2022-05-05
Switch to dockerhub image for bitnami

bpradipt:
The bitnami images will only be available from dockerhub and gcr going
forward
Fixes: #34


fidencio (approved):
lgtm, thanks @bpradipt!

PR #37, opened 2022-05-10
Update the bundle and example manifests to be in sync with latest CRD changes

### LINKED ISSUE #36: The bundle manifests and deploy.yaml are not in sync with latest CRD changes
**Describe the bug**
The bundle manifests and deploy.yaml are not in sync with latest CRD changes. Consequently, it's not possible to create the CR using `deploy/deploy.yaml`

**To Reproduce**
1. kubectl apply -f deploy/deploy.yaml
2. kubectl apply -f config/samples/ccruntime.yaml

This will fail

**Describe the results you expected**
Successful creation of CR

**Describe the results you received:**
Error indicating schema mismatch
 
**Additional context**
This issue is due to usage of old schema in the deploy.yaml

bpradipt:
Fixes #36


fidencio (approved):
lgtm, thanks @bpradipt!

This is something easy to miss, by the way, and when we start thinking about CI ... that's a check we can add to avoid such issues in the future.

Issue #50, opened 2022-07-07
pre-install / post-uninstall daemonsets are not started on reinstall


fidencio:
**Describe the bug**
After deleting the `ccruntime.yaml` CRD, and applying it again, the pre-instal / post-uninstall daemonsets are never started

**To Reproduce**
1. Deploy the operator
2. Deploy the CRD
3. Delete the CRD
4. Manually cleanup the daemonsets left behind (see: https://github.com/confidential-containers/operator/issues/49)
5. Re-deploy the CRD

**Describe the results you expected**
The pre-install / post-uninstall daemonsets should be running as they were during the first installation

**Describe the results you received:**
The pre-install / post-uninstall daemonsets are never started

Issue #54, opened 2022-07-18
label of pre/post daemonsets are not removed after uninstall


jensfr:
**Describe the bug**
(Un)InstallDone labels should be removed after a daemonset ran to completion.
Otherwise on the next run the controller finds the labels and does nothing.

**To Reproduce**
create a ccruntime example CR, remove it and then create it again

**Describe the results you expected**
Files should be reliably deployed on the nodes on the second installatin

**Describe the results you received:**
nothing happens because the controller code sees labels and considers
the task as done
 
**Additional context**

PR #55, opened 2022-07-20
allow re-installation by deleting labels and preinstall daemonset

jensfr:
This contains two patches. The first one adds code to delete pre-install and post-uninstall
done labels. The second one deletes the pre-install daemon set so that it will be started
again when a new CR is created. 

Fixes: #50 #54

### DIFF
--- a/controllers/ccruntime_controller.go
+++ b/controllers/ccruntime_controller.go
@@ -745,12 +745,17 @@ func (r *CcRuntimeReconciler) SetupWithManager(mgr ctrl.Manager) error {
 		Complete(r)
 }
 func (r *CcRuntimeReconciler) deleteUninstallDaemonsets() (ctrl.Result, error) {
-	ds := r.processDaemonset(InstallOperation)
+	ds := r.processDaemonset(UninstallOperation)
 	result, err := r.deleteDaemonset(ds)
 	if err != nil {
 		return result, err
 	}
 	ds = r.makeHookDaemonset(PostUninstallOperation)
+	result, err = r.deleteDaemonset(ds)
+	if err != nil {
+		return result, err
+	}
+	ds = r.makeHookDaemonset(PreInstallOperation)
 	result, err = r.deleteDaemonset(ds)
 	if err != nil {
 		return result, err
--- a/controllers/ccruntime_controller.go
+++ b/controllers/ccruntime_controller.go
@@ -228,6 +228,32 @@ func (r *CcRuntimeReconciler) processCcRuntimeDeleteRequest() (ctrl.Result, erro
 						return result, err
 					}
 					result, err = r.deleteUninstallDaemonsets()
+					prepostLabels := map[string]string{}
+					if r.ccRuntime.Spec.Config.PreInstall.Image != "" {
+						prepostLabels[PreInstallDoneLabel[0]] = PreInstallDoneLabel[1]
+					}
+					if r.ccRuntime.Spec.Config.PostUninstall.Image != "" {
+						prepostLabels[PostUninstallDoneLabel[0]] = PostUninstallDoneLabel[1]
+
+					}
+					err, nodes := r.getNodesWithLabels(prepostLabels)
+					if err != nil {
+						r.Log.Error(err, "an error occured when getting the list of nodes from which we want to"+
+							"remove preinstall/postuninstall labels")
+						return ctrl.Result{}, err
+					}
+					postuninstalledNodes := len(nodes.Items)
+					if !allNodesDone(postuninstalledNodes, r) {
+						return ctrl.Result{Requeue: true}, nil
+					} else {
+						if postuninstalledNodes > 0 {
+							result, err = r.removeNodeLabels(nodes)
+							if err != nil {
+								r.Log.Error(err, "removing the labels from nodes failed")
+								return ctrl.Result{}, err
+							}
+						}
+					}
 				}
 			}
 			return result, err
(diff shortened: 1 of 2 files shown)


fidencio:
@jensfr, in order to speed up the tests, would you happen to have an image already built with those fixes?

jensfr:
@fidencio I just pushed quay.io/jensfr/confidential-containers-operator:pull-55 which has these fixes. Thanks for reviewing and testing!

fidencio:
> @fidencio I just pushed quay.io/jensfr/confidential-containers-operator:pull-55 which has these fixes. Thanks for reviewing and testing!

Almost there, @jensfr.  I've noticed that the runtime classes are not created on reinstall.

jensfr:
odd, that never failed in my tests. I'll try to reproduce, thx!

controllers/ccruntime_controller.go
```diff
@@ -228,6 +228,32 @@ func (r *CcRuntimeReconciler) processCcRuntimeDeleteRequest() (ctrl.Result, erro
...
+					}
+					err, nodes := r.getNodesWithLabels(prepostLabels)
+					if err != nil {
+						r.Log.Error(err, "an error occured when getting the list of nodes from which we want to"+
+							"remove preinstall/postuninstall labels")
+						return ctrl.Result{}, err
+					}
+					postuninstalledNodes := len(nodes.Items)
+					if !allNodesDone(postuninstalledNodes, r) {
+						return ctrl.Result{Requeue: true, RequeueAfter: time.Second * 5}, err
```
bpradipt, line 247: `RequeueAfter` is not required since for any errors, by default the request is queued with exponential backoff. It could be changed to `return ctrl.Result{}, err`

controllers/ccruntime_controller.go
```diff
@@ -228,6 +228,32 @@ func (r *CcRuntimeReconciler) processCcRuntimeDeleteRequest() (ctrl.Result, erro
...
+					} else {
+						if postuninstalledNodes > 0 {
+							result, err = r.removeNodeLabels(nodes)
+							if err != nil {
+								r.Log.Error(err, "removing the labels from nodes failed")
+								return ctrl.Result{Requeue: true, RequeueAfter: time.Second * 5}, err
+							}
```
bpradipt, line 254: Same comment as above w.r.to RequeueAfter usage

controllers/ccruntime_controller.go
```diff
@@ -719,7 +745,7 @@ func (r *CcRuntimeReconciler) SetupWithManager(mgr ctrl.Manager) error {
 		Complete(r)
 }
 func (r *CcRuntimeReconciler) deleteUninstallDaemonsets() (ctrl.Result, error) {
```
bpradipt, line 747: note: in a separate PR, it might be worth looking at renaming this method since it is not restricted to Uninstall DS :-)

controllers/ccruntime_controller.go
```diff
@@ -847,3 +878,33 @@ func (r *CcRuntimeReconciler) makeHookDaemonset(operation DaemonOperation) *apps
...
+		r.Log.Info("Couldn't get nodes client")
+		return ctrl.Result{}, err
+	}
+
+	for _, node := range nodesList.Items {
+		nodeLabels := node.GetLabels()
+		for k, v := range nodeLabels {
+			if k == PostUninstallDoneLabel[0] && v == PostUninstallDoneLabel[1] ||
+				k == PreInstallDoneLabel[0] && v == PreInstallDoneLabel[1] ||
+				k == "katacontainers.io/kata-runtime" && v == "cleanup" {
```
bpradipt, line 894: How about defining the above runtime specific label also in common.go ? 
Keeping all the required labels in one place for easier reference

I want to suggest a style change with an example. It will make the code easier to read when adding new labels for different runtimes. 
```
if val, ok := nodeLabels[PreInstallDoneLabel[0]]; ok && val == PreInstallDoneLabel[1] {
          delete(nodeLabels, PreInstallDoneLabel[0])
 }

if val, ok := nodeLabels[PostUninstallDoneLabel[0]]; ok && val == PostUninstallDoneLabel[1] {
           delete(nodeLabels, PostUninstallDoneLabel[0])
}

if val, ok := nodeLabels[KataRuntimeLabel[0]]; ok && val == KataRuntimeLabel[1] {
          delete(nodeLabels, KataRuntimeLabel[0])
}
```

bpradipt:
I successfully tested the code changes in my setup and I could see the daemonsets being cleaned up during uninstall and recreated for reinstall.

controllers/ccruntime_controller.go
jensfr: good point Pradipta, this wasn't clear to me but it makes sense. I'll change it here and in other places

controllers/ccruntime_controller.go
jensfr: yes, true. I'll open an issue for it

controllers/ccruntime_controller.go
jensfr: good suggestion, I'll adapt it accordingly!

controllers/ccruntime_controller.go
jensfr: I changed it, but left the `Requeue: true` in because in this case we definitely want to requeue but err is not set (allNodesDone doesn't set it, the value comes from a previous call which is already checked. 
So now it is `return ctrl.Result{Requeue: true}, nil`. 

I used your suggestion in the other case a bit further down. There are more places in the code that need cleanup but that's out of scope for this PR.

jensfr:
@bpradipt I addressed your comments and tested it. New image is quay.io/jensfr/confidential-containers-operator:pull-55-3

bpradipt (approved):
/lgtm
Thanks @jensfr

Issue #58, opened 2022-08-01
uninstall hangs when no pre-/postuninstall daemon sets used


jensfr:
**Describe the bug**
Uninstall never finishes when no pre-/postuninstall  daemon sets are used. 
oc command never returns.

**To Reproduce**
oc create -f ccruntime-sample.yaml
wait until finished
oc delete -f ccruntime-sample.yaml
note that oc never returns

**Describe the results you expected**
Uninstallation should finish and oc command should return.

**Describe the results you received:**
oc did not return and the installation was never finished
 
**Additional context**
Reported by @snir911 
The finalizer of the ccruntime object is never removed.
When removing it manually the uninstallation is finished and oc returns.

PR #59, opened 2022-08-01
Update ccruntime after finalizer removed

jensfr:
We need to update the CCRuntime CR after we have removed
the finalizer. When no postuninstall DaemonSet was used we
return values that mean we don't update the CR instance.

Fix this by updating the status before returning.

Fixes: #58 
Reported-by: Snir Sheriber <ssheribe@redhat.com>


bpradipt (approved):
/lgtm
Thanks @jensfr

Issue #62, opened 2022-08-25
RuntimeClass is not being set on k8s 1.25


snir911:
**Describe the bug**
The operator is silently failing to set the runtimeclass objects on **1.25** k8s, although finalizers are added as expected

**To Reproduce**
1. deploy the operator
2. deploy the ccruntime CR

**Describe the results you expected**
kata-clh,kata-qemu,kata runtimeclass would be added

**Describe the results you received:**
`kubectl get runtimeclass` returns nothing
 
**Additional context**
Tested with peer-pods

fidencio:
I faced the very same issue with confidential containers.

bpradipt:
The issue is due to using deprecated api library in the code `nodeapi k8s.io/api/node/v1beta1`
@fidencio tried the fix in @snir911 environment and it worked. Can you try this in your environment 
```
kubectl set image deployment/cc-operator-controller-manager  -n confidential-containers-system manager=quay.io/bpradipt/cc-operator:latest
```

bpradipt:
There are additional changes w.r.to latest k8s apis. I'll do some more investigations before raising a PR

PR #65, opened 2022-08-26
This patch series upgrades the operator-sdk and associated dependencies

bpradipt:
This series also fixes: #62

Tested on kubernetes version 1.24 and 1.25


bundle/manifests/cc-operator.clusterserviceversion.yaml
```diff
@@ -12,21 +12,16 @@ metadata:
...
+              "matchLabels": {
+                "node-role.kubernetes.io/worker": ""
+              }
+            },
             "config": {
               "cleanupCmd": [
                 "/opt/kata-artifacts/scripts/kata-deploy.sh",
                 "reset"
               ],
-              "cleanupNodeSelector": [
```
jensfr, line 20: this is deleted because it's not in use any more?

jensfr: looks like we have it hard coded in a few places. I guess that's okay, we'll fix it in a follow up

jensfr (approved):
looks good to me, thank  you @bpradipt

jensfr:
tested it successfully, thanks for the quick fix!

bundle/manifests/cc-operator.clusterserviceversion.yaml
bpradipt: Hmm, `make bundle` synced with the manifests under `config`. Let me just cross-check once to be sure. Thanks for catching this.

bpradipt: @jensfr as you rightly mentioned the `cleanupNodeSelector` is no longer used and hence removed from bundle.

PR #63, opened 2022-08-25
ci: bootstrap the test env and install the operator

wainersm:
Adds ansible playbooks to setup the test environment. Kubernetes will be
deployed with kubeadm.

The cluster/up.sh is used to bring up the cluster. Afterwards run
build_install_operator.sh to build and install the operator and
ccruntime. Notice it pushes the operator's controller image to a docker
registry running locally so we avoid pushing/pulling to/from quay.io.

I tested in a Fedora 34 x86_64 machine in a Vagrant VM (see the
Vagrantfile added in this commit). If you wish to run likewise:

$ cd tests/e2e
$ vagrant up tests-e2e-ubuntu2004

Signed-off-by: Wainer dos Santos Moschetta <wainersm@redhat.com>

PR #70, opened 2022-08-31
build: ignore crd not found in uninstall/undeploy targets

jensfr:
uninstall and undeploy targets both want to delete the CCRuntimeConfig
CRD. When we call these targets one after another the second one will
always fail. This is because we add 'ignore-not-found=false' when we
call `kubectl delete`.

It is safe to set `ignore-not-found=true` in the undelete/undeploy case
because if a resource is not there we don't have to delete it.

Signed-off-by: Jens Freimann <jfreimann@redhat.com>

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -126,7 +126,7 @@ docker-push: ## Push docker image with the manager.
 ##@ Deployment
 
 ifndef ignore-not-found
-  ignore-not-found = false
+  ignore-not-found = true
 endif
 
 .PHONY: install


wainersm (approved):
Hi @jensfr !

I applied this changed and I no longer see the error:
```
test -s /home/vagrant/src/confidential-containers/operator/bin/controller-gen || GOBIN=/home/vagrant/src/confidential-containers/operator/bin go install sigs.k8s.io/controller-tools/cmd/controller-gen@v0.9.2
/home/vagrant/src/confidential-containers/operator/bin/controller-gen rbac:roleName=manager-role crd webhook paths="./..." output:crd:artifacts:config=config/crd/bases
/home/vagrant/src/confidential-containers/operator/bin/kustomize build config/crd | kubectl delete --ignore-not-found=true -f -
customresourcedefinition.apiextensions.k8s.io "ccruntimes.confidentialcontainers.org" deleted
/home/vagrant/src/confidential-containers/operator/bin/kustomize build config/default | kubectl delete --ignore-not-found=true -f -
namespace "confidential-containers-system" deleted
serviceaccount "cc-operator-controller-manager" deleted
role.rbac.authorization.k8s.io "cc-operator-leader-election-role" deleted
clusterrole.rbac.authorization.k8s.io "cc-operator-manager-role" deleted
clusterrole.rbac.authorization.k8s.io "cc-operator-metrics-reader" deleted
clusterrole.rbac.authorization.k8s.io "cc-operator-proxy-role" deleted
rolebinding.rbac.authorization.k8s.io "cc-operator-leader-election-rolebinding" deleted
clusterrolebinding.rbac.authorization.k8s.io "cc-operator-manager-rolebinding" deleted
clusterrolebinding.rbac.authorization.k8s.io "cc-operator-proxy-rolebinding" deleted
configmap "cc-operator-manager-config" deleted
service "cc-operator-controller-manager-metrics-service" deleted
deployment.apps "cc-operator-controller-manager" deleted
```
Thanks!

bpradipt:
@jensfr I'm wondering if we need to change the default from the operator-sdk provided one, since the Makefile targets explicitly mentions about using `ignore-not-found=true` if required:
https://github.com/confidential-containers/operator/blob/main/Makefile#L137
https://github.com/confidential-containers/operator/blob/main/Makefile#L146

Should this behaviour be instead documented leaving the default as-is?

jensfr:
> @jensfr I'm wondering if we need to change the default from the operator-sdk provided one, since the Makefile targets explicitly mentions about using `ignore-not-found=true` if required: https://github.com/confidential-containers/operator/blob/main/Makefile#L137 https://github.com/confidential-containers/operator/blob/main/Makefile#L146
> 
> Should this behaviour be instead documented leaving the default as-is?

You're absolutely correct. @wainersm this PR isn't even needed. Just call it with `make ignore-not-found=true uninstall undeploy` and the effect should be the same.

bpradipt:
> > > @jensfr I'm wondering if we need to change the default from the operator-sdk provided one, since the Makefile targets explicitly mentions about using `ignore-not-found=true` if required: https://github.com/confidential-containers/operator/blob/main/Makefile#L137 https://github.com/confidential-containers/operator/blob/main/Makefile#L146
> > > Should this behaviour be instead documented leaving the default as-is?
> > 
> > 
> > You're absolutely correct. @wainersm this PR isn't even needed. Just call it with `make ignore-not-found=true uninstall undeploy` and the effect should be the same.
> 
> how end user will uninstall the operator? If not using those make targets, it would be better to change on CI.

`make help` describe the options that are available.

What you can do is run these commands in the CI.
`make uninstall ignore-not-found=true`
`make undeploy ignore-not-found=true`

jensfr:
Well, I'd say end users create a subscription via OLM. But that's probably a discussion for https://github.com/confidential-containers/operator/pull/63 
The two make targets are the correct ones two get rid of everything

jensfr:
closed because the change is not needed, instead one can set the ignore-not-found variable

PR #72, opened 2022-09-05
bundle,config: Update pre/post image

fidencio:
Let's update the pre/post hooks image to point to the latest version of
the container-engine-for-cc-payload.

The changes in the container-engine-for-cc-payload are:
* Removing `/opt/confidential-containers` as part of the uninstall
  target.


bpradipt (approved):
/lgtm

PR #73, opened 2022-09-05
bundle,config: Update Kata Containers payload

fidencio:
Let's update the Kata Containers payload to its latest build, which is
based on the commit f5a65223989b090c91e94faf933ce256fc6e6c9b.

The most significant changes which are part of this new image are:
* TDX support

### DIFF
--- a/bundle/manifests/cc-operator.clusterserviceversion.yaml
+++ b/bundle/manifests/cc-operator.clusterserviceversion.yaml
@@ -106,7 +106,7 @@ metadata:
               ],
               "payloadImage": "quay.io/confidential-containers/runtime-payload:kata-containers-f5a65223989b090c91e94faf933ce256fc6e6c9b",
               "postUninstall": {
-                "image": "quay.io/confidential-containers/container-engine-for-cc-payload:2022090517241662391458",
+                "image": "quay.io/confidential-containers/container-engine-for-cc-payload:2022070719511657216287",
                 "volumeMounts": [
                   {
                     "mountPath": "/opt/confidential-containers/",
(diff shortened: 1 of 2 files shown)


bpradipt:
@fidencio are there plans to have different payloads for tdx, sev etc ?
In that case we can look at ccruntime manifest for tdx ,sev etc (eg. ccruntime-tdx.yaml, ccruntime-sev.yaml) and you can rename the file in this PR.

fidencio:
> @fidencio are there plans to have different payloads for tdx, sev etc ? In that case we can look at ccruntime manifest for tdx ,sev etc (eg. ccruntime-tdx.yaml, ccruntime-sev.yaml) and you can rename the file in this PR.

Not right now, Pradipta.  Right now, at least for the first release, we're going with a single payload that will deploy both non-TEE, TDX, and SEV.

fidencio:
> > @fidencio are there plans to have different payloads for tdx, sev etc ? In that case we can look at ccruntime manifest for tdx ,sev etc (eg. ccruntime-tdx.yaml, ccruntime-sev.yaml) and you can rename the file in this PR.
> 
> Not right now, Pradipta. Right now, at least for the first release, we're going with a single payload that will deploy both non-TEE, TDX, and SEV.

Let me expand a little bit on this.  The whole image with all the TEEs bits will be around 600MB, 650MB at the worst case scenario, while the image just for non-TEE support is around 450MB, and the majority of the size comes from the "CentOS 7" base that we need to do operations like using `kubectl` and doing a `systemctl restart` from inside the container.

So, I think it's easier to just have everything inside the same image and, later on, rely on NFD to install whatever is needed in a specific cluster, than splitting the images and increasing the load on the Kata Containers payload release size / Operator.

Plus, if someone has a non-homogeneous / mixed setup (which I can see happening in the future), a single image would be better to rely on a single image for the whole fleet.

bpradipt (approved):
/lgtm
Thanks @fidencio .

bpradipt:
/hold
Putting the merge on hold for @wainersm to test this PR via CI

wainersm:
let me have a look at the failed job

fidencio:
I ended up reverting 3267390968e83d9722e46e461d4bebbb8d0827ca as it's been causing pre-install to hang.
I will raise another PR updating that one after debugging it.

PR #75, opened 2022-09-06
ci: allow to run tests for different runtimeclasses

wainersm:
Added the -r RUNTIMECLASS option to both run-local.sh and
tests_runner.sh so that CI jobs will be able to specify the RUNTIMECLASS
to be used on tests.

For example, the following to run with kata-clh and kata-qemu-tdx
respectively:

$ ./run-local.sh -r kata-clh
$ ./run-local.sh -r kata-qemu-tdx

Signed-off-by: Wainer dos Santos Moschetta <wainersm@redhat.com>

### DIFF
--- a/tests/e2e/run-local.sh
+++ b/tests/e2e/run-local.sh
@@ -10,22 +10,30 @@ set -o pipefail
 
 script_dir="$(dirname "$(readlink -f "$0")")"
 
+runtimeclass=""
 undo="false"
 
 usage() {
 	cat <<-EOF
-	$0: run the end-to-end tests on local host.
-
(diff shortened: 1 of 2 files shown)


fitzthum (approved):
Ah great I was literally just wondering about this. It looks like the idea is to always use the `kata` runtime class but then redirect the config file to the appropriate one. Technically this doesn't quite cover the containerd config file in the way that it will be used in production, but I think it's probably fine. 

Also this way we don't have to modify tests that expect the `kata` runtime class.

fidencio (approved):
lgtm, amazing work @wainersm!

PR #76, opened 2022-09-06
bundle,config: Update pre-install / post-uninstall image

fidencio:
Thanks to our new CI, we've found that the previous update of the pre-install / post-uninstall image actually introduced a regression in the Operator, as the script would fail trying to remove a non-empty `/opt/confidential-containers` directory.

The main changes in this version of the image are:
* Fixes a bug on uninstall, where we'd try to remove containerd while
  its binary was still being used
* Includes more debug messages
* Uses a commit id as the image tag, instead of the time the image was
  built, in order to make it easier to link an image to the code.


wainersm:
@fidencio not the first time I see that failure to build kustomize. Any idea what is the problem? I wonder if this is an intermittent failure not related with the PR changes.

bpradipt (approved):
/lgtm

PR #79, opened 2022-09-08
CI: add support to CentOS

wainersm:
The CI job for Intel TDX should run on CentOS Stream 8, therefore, it is required the scripts (mostly the Ansible playbooks) to support that distro.


wainersm:
Hi @fidencio @arronwy , I made this changes for the TDX CI. I tested it locally on a Vagrant VM and it installed everything up to the point of the operator, then it failed because the VM hasn't enough disk. I'm confident it should work fine. So could you please review and test it on your environment?

fidencio:
> /test-kata-qemu-tdx

I am sincerely not sure if this has any chance to work, but it allows me to also check the new job added on the TDX CI side.

tests/e2e/ansible/install_docker.yml
```diff
@@ -0,0 +1,69 @@
+# Copyright 2022 Red Hat
```
bpradipt, line 1: nit: For all new code, let's follow CNCF [guidelines](https://github.com/cncf/foundation/blob/main/copyright-notices.md) 

and use something like this:
```
Copyright Confidential Containers Contributors
```

We need to do this for all the code, but let's start with this one :-)

tests/e2e/ansible/install_docker.yml
```diff
@@ -0,0 +1,69 @@
...
+      apt_repository:
+        repo: "deb [arch=amd64 signed-by=/etc/apt/trusted.gpg.d/docker.gpg] https://download.docker.com/linux/ubuntu {{ ansible_lsb.codename }} stable"
+        filename: docker
+        update_cache: yes
+        state: present
+    - name: Install docker packages
+      package:
+        name:
+          - docker-ce
+          - docker-ce-cli
```
bpradipt, line 40: `containerd.io` need to be installed ? 
Ref: https://docs.docker.com/engine/install/ubuntu/

tests/e2e/ansible/install_docker.yml
```diff
@@ -0,0 +1,69 @@
...
+        state: present
+    - name: Add docker yum repo
+      shell: yum-config-manager --add-repo https://download.docker.com/linux/centos/docker-ce.repo
+      args:
+        creates: /etc/yum.repos.d/docker-ce.repo
+    - name: Install docker packages
+      dnf:
+        name:
+          - docker-ce
+          - docker-ce-cli
```
bpradipt, line 62: same comment on `containerd.io`
Ref: https://docs.docker.com/engine/install/rhel/

tests/e2e/ansible/install_docker.yml
fidencio: According to the CentOS guide, yes: https://docs.docker.com/engine/install/centos/

fidencio:
Let me just do one quick test by closing and re-opening this PR.

tests/e2e/run-local.sh
```diff
@@ -69,19 +69,21 @@ main() {
...
+	sudo -E PATH="$PATH" bash -c './cluster/up.sh'
 	export KUBECONFIG=/etc/kubernetes/admin.conf
 
 	echo "INFO: Build and install the operator"
-	sudo -E PATH="$PATH" ./operator.sh
+	sudo -E PATH="$PATH" bash -c './operator.sh'
 
 	echo "INFO: Run tests"
-	cmd="sudo -E PATH=\"$PATH\" ./tests_runner.sh"
+	cmd="sudo -E PATH=\"$PATH\" bash -c './tests_runner.sh'"
```
arronwy, line 86: `'` here may lead `-r $runtimeclass` failed to append to tests_runner.sh as parameter

arronwy (commented):
I found the `runtimeClassName` defined in Kata tests cases are `kata` by default, in our `tests_runner.sh` script, we may also need update them with the `runtimeclass` parameter like:
```
sed -i  "s/runtimeClassName: kata/runtimeClassName: $1/g"  fixtures/*.yaml
```

tests/e2e/ansible/install_docker.yml
wainersm: @bpradipt I read that before but I misunderstood the doc. thanks for the clarification, I will fix them on this PR.

tests/e2e/ansible/install_docker.yml
wainersm: yes, it does. I didn't notice that because in `tests/e2e/ansible/install_containerd.yml` I end up installing containerd. But it makes sense to get it installed in install_docker.yml as I am trying to make those playbook self-contained.

Good catch. Gonna fix it.

tests/e2e/run-local.sh
wainersm: Good catch! Gonna fix it!

wainersm:
Update the PR with:
- fixed copyright in install_docker.sh (@bpradipt )
- install containerd.io in install_docker.sh (@bpradipt , @fidencio )
- fixed `$cmd` build (@arronwy )

wainersm:
> I found the `runtimeClassName` defined in Kata tests cases are `kata` by default, in our `tests_runner.sh` script, we may also need update them with the `runtimeclass` parameter like:
> 
> ```
> sed -i  "s/runtimeClassName: kata/runtimeClassName: $1/g"  fixtures/*.yaml
> ```

hi @arronwy,
 
Yes, I addressed that with a dirty hack: see https://github.com/confidential-containers/operator/blob/main/tests/e2e/tests_runner.sh#L70 . Recently @fitzthum sent the PR #82 that reverts my changes but doesn't solve the problem.

Yesterday  I started adapt the tests so that they will recognize a environment variable called RUNTIMECLASS to generate the pod config file with the correct runtimeClassName, therefore not needing to run the `sed` command you proposed on PR #85 . I will send that today. My other concern was the tests having to change kata's configuration.toml that you also address on #85, I will review that still today too.

Besides those 'details' on the tests: the code that I proposed on this PR to setup the environment on CentOS is right and enough?

arronwy:
> > I found the `runtimeClassName` defined in Kata tests cases are `kata` by default, in our `tests_runner.sh` script, we may also need update them with the `runtimeclass` parameter like:
> > ```
> > sed -i  "s/runtimeClassName: kata/runtimeClassName: $1/g"  fixtures/*.yaml
> > ```
> 
> hi @arronwy,
> 
> Yes, I addressed that with a dirty hack: see https://github.com/confidential-containers/operator/blob/main/tests/e2e/tests_runner.sh#L70 . Recently @fitzthum sent the PR #82 that reverts my changes but doesn't solve the problem.
> 
> Yesterday I started adapt the tests so that they will recognize a environment variable called RUNTIMECLASS to generate the pod config file with the correct runtimeClassName, therefore not needing to run the `sed` command you proposed on PR #85 . I will send that today. My other concern was the tests having to change kata's configuration.toml that you also address on #85, I will review that still today too.

Yes, thanks, the changes only modify the code from tests repo.

> 
> Besides those 'details' on the tests: the code that I proposed on this PR to setup the environment on CentOS is right and enough?

Yes, with this PR, CI jobs runs ok on CentOS

wainersm:
Just rebased the PR to run the tests to ensure this PR doesn't break on Ubuntu.

/test

sameo (approved):

arronwy:
> @arronwy I'd like to merge this if you agree. If yes, please give the approval :)

sure, thanks @wainersm , TDX CI also passed.

arronwy (approved):

PR #80, opened 2022-09-08
config: add kata-qemu-sev runtimeclass

Alex-Carter01:
Adding runtimeclass for sev. 

Depends on sev kata PRs:

- Packaging for initrd: https://github.com/kata-containers/kata-containers/pull/5120
- Packaging for OVMF: https://github.com/kata-containers/kata-containers/pull/5026
- Packaging for SEV-capable kernel + kernel module https://github.com/kata-containers/kata-containers/pull/5038
- Add agent config to rootfs: https://github.com/kata-containers/kata-containers/pull/5024
- Use depmod in rootfs builder on kernel module(s): https://github.com/kata-containers/kata-containers/pull/5127
- Kata-deploy add kata-qemu-sev shim/runtimeclas https://github.com/kata-containers/kata-containers/pull/5130

Integration testing of all of the SEV components has been successful. Made a cc bundle with all of the PRs added to base CCv0, converted to a docker container via the kata-deploy-cc docker script, and deployed to quay. Was able to run a pod after being deployed with the operator.

@fidencio @fitzthum @wainersm 

Signed-off-by: Alex Carter <Alex.Carter@ibm.com>

PR #82, opened 2022-09-16
Switch between tests based on runtime class 

fitzthum:
Rather than using the RUNTIMECLASS parameter to modify the contained config, use it to switch between different test files. These tests will then use the appropriate runtime class. The runtime class is a good way to identify different tests because it includes the hypervisor and the hardware that are supported.

This also adds the SEV tests that @ryansavino is working on. See https://github.com/kata-containers/tests/pull/5130

Depends on https://github.com/confidential-containers/operator/pull/80 for  the runtime class.

At some point in the future we might want a more robust way to link test file names to rurntime classes. It's also a bit annoying that we have to specify the names of each of the tests in this repo. That will mean we need extra changes when we push new tests to the tests repo.

@wainersm

### DIFF
--- a/tests/e2e/tests_runner.sh
+++ b/tests/e2e/tests_runner.sh
@@ -73,21 +73,8 @@ usage() {
 	EOF
 }
 
-main() {
-
-	parse_args $@
-
-	clone_kata_tests
-
-	cd "${tests_repo_dir}/integration/kubernetes/confidential"
-
-	# Test scripts rely on kata-runtime so it should be reacheable on PATH.
-	# Re-export PATH is error prone as some calls to kata-runtime use sudo,
-	# so let's create a symlink.
-	ln -sf /opt/confidential-containers/bin/kata-runtime \
-		/usr/local/bin/kata-runtime
-
-	# Run tests.
+# tests for CC with QEMU and no specific hardware support
+run_kata_qemu_tests() {
 
 	# Results for agent_image.bats:
 	#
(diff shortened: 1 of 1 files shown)


tests/e2e/tests_runner.sh
```diff
@@ -112,6 +100,49 @@ main() {
 		"agent_image.bats" \
 		"agent_image_encrypted.bats" \
 		"${script_dir}/operator_tests.bats"
+
+}
+
+# Tests for CC with QEMU on SEV HW
+run_kata_qemu_sev_tests() {
+	local tests_passing="Test SEV unencrypted container launch success"
```
wainersm, line 108: If all tests are passing then you don't need to list them individually. I made that way on my initial implementation because two of the non-tee tests were failing.

tests/e2e/tests_runner.sh
```diff
@@ -112,6 +100,49 @@ main() {
...
+		kata-qemu)
+			echo "INFO: Running kata-qemu tests"
+			run_kata_qemu_tests
+			;;
+		kata-qemu-sev)
+			echo "INFO: Running kata-qemu-sev tests"
+			run_kata_qemu_sev_tests
+			;;
+		*)
+			echo "ERROR: unknown runtimeclass"
```
wainersm, line 142: actually unknown tests for runtimeclass X.

tests/e2e/tests_runner.sh
```diff
@@ -112,6 +100,49 @@ main() {
...
+	case $runtimeclass in
+		kata-qemu)
+			echo "INFO: Running kata-qemu tests"
+			run_kata_qemu_tests
+			;;
+		kata-qemu-sev)
+			echo "INFO: Running kata-qemu-sev tests"
+			run_kata_qemu_sev_tests
+			;;
+		*)
```
wainersm, line 141: @arronwy we will need to add an entry for TDX here. IIUC we will be running the non-TEE tests (i.e. same as in run_kata_qemu_tests()) for TDX but tweaking the runtimeclass. Is that what you want?

Moreover, if we merge it as is it will break your tests you have made manually. Is it okay to merge it anyway then we open a follow up PR to add the TDX tests in?

arronwy: > @arronwy we will need to add an entry for TDX here. IIUC we will be running the non-TEE tests (i.e. same as in run_kata_qemu_tests()) for TDX but tweaking the runtimeclass. Is that what you want?

Yes, non-TEE tests can be the common test cases for all the runtimeclasses, we can set correct  `RUNTIMECLASS`  afterthis PR: https://github.com/kata-containers/tests/pull/5131 is merged.

> Moreover, if we merge it as is it will break your tests you have made manually. Is it okay to merge it anyway then we open a follow up PR to add the TDX tests in?

It's ok for me to merge this PR and a follow up PR for other runtimeclasses with default non-TEE tests.

wainersm: > > @arronwy we will need to add an entry for TDX here. IIUC we will be running the non-TEE tests (i.e. same as in run_kata_qemu_tests()) for TDX but tweaking the runtimeclass. Is that what you want?
> 
> Yes, non-TEE tests can be the common test cases for all the runtimeclasses, we can set correct `RUNTIMECLASS` afterthis PR: [kata-containers/tests#5131](https://github.com/kata-containers/tests/pull/5131) is merged.
>

Notice that some tests still write to kata's configuration.toml and I don't know how to avoid that.
 
> > Moreover, if we merge it as is it will break your tests you have made manually. Is it okay to merge it anyway then we open a follow up PR to add the TDX tests in?
> 
> It's ok for me to merge this PR and a follow up PR for other runtimeclasses with default non-TEE tests.

Cool. so I will merge it.

wainersm (approved):
Thanks @fitzthum

PR #84, opened 2022-09-19
Install bats from upstream 

### LINKED ISSUE #83: Operator CI installing old bats version
The operator CI lists `bats` as a test dependency (see `install_test_deps.yml`), which causes `bats` to be installed via apt. Unfortunately apt tracks an old version of bats, which does not match the one used in the kata CI and does not support the SEV tests. The kata CI uses a `.ci/install_bats.sh` script to install bats from upstream. Maybe we should do this in the operator CI as well? 

cc @wainersm @ryansavino

fitzthum:
Fixes #83. Based on [this script](https://github.com/kata-containers/tests/blob/main/.ci/install_bats.sh)

Idk how to use ansible so maybe there is a much nicer way to do this.

@wainersm  @ryansavino

### DIFF
--- a/tests/e2e/ansible/install_test_deps.yml
+++ b/tests/e2e/ansible/install_test_deps.yml
@@ -7,7 +7,6 @@
     - name: Install test dependencies
       package:
         name:
-          - bats
           - jq
         state: present
     - block:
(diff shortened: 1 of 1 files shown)


wainersm:
@fitzthum In my PR to add CentOS support (#79) I've a snippet of code that installs bats from source unless it is already available in the system. Maybe I can generalize that.

wainersm:
btw, here is the code -> https://github.com/confidential-containers/operator/pull/79/commits/c22e77eb09740f6a7f036b71e6f19f4bba339219#diff-87d51b490c4353f16b47e9bfa4d94a545304e9ab1009580bce56cd37ab0ad7abR12

fitzthum:
Yeah fine to use the CentOS stuff as long as that is getting in soon. Looks quite similar (but better) to this.

wainersm:
Superseded by https://github.com/confidential-containers/operator/pull/89

PR #89, opened 2022-09-21
ci: install bats from source

### LINKED ISSUE #83: Operator CI installing old bats version

wainersm:
The bats version on Ubuntu 20.04 don't have all features we need (e.g. setup_file()), so let's build from source. On CentOS it is already built and installed from source, so just enabled it to Ubuntu as well.

Fixes #83
Signed-off-by: Wainer dos Santos Moschetta <wainersm@redhat.com>

### DIFF
--- a/tests/e2e/ansible/group_vars/all
+++ b/tests/e2e/ansible/group_vars/all
@@ -22,7 +22,6 @@ kubeadm_pkgs:
 k8s_version: v1.24.0
 test_pkgs:
   ubuntu:
-    - bats
     - jq
   centos:
     - jq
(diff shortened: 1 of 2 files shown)


fitzthum (approved):
Looks great

PR #91, opened 2022-09-23
api: accept enclave-cc runtimeName

mythi:
Currently, only kata is allowed as runtimeName in the API. The field is copied to CcRuntimeStatus but other than that it's unused.

The first non-kata runtime is nicknamed 'enclave-cc'. Add support for it as a runtimeName identifier.

### DIFF
--- a/config/crd/bases/confidentialcontainers.org_ccruntimes.yaml
+++ b/config/crd/bases/confidentialcontainers.org_ccruntimes.yaml
@@ -5556,6 +5556,7 @@ spec:
               runtimeName:
                 enum:
                 - kata
+                - enclave-cc
                 type: string
             required:
             - config
@@ -5633,6 +5634,7 @@ spec:
                 description: Cc Runtime Name
                 enum:
                 - kata
+                - enclave-cc
                 type: string
               totalNodesCount:
                 description: TotalNodesCounts is the total number of worker nodes
--- a/api/v1beta1/ccruntime_types.go
+++ b/api/v1beta1/ccruntime_types.go
@@ -24,15 +24,8 @@ import (
 // EDIT THIS FILE!  THIS IS SCAFFOLDING FOR YOU TO OWN!
 // NOTE: json tags are required.  Any new fields you add must have json tags for the fields to be serialized.
 
-// +kubebuilder:validation:Enum=kata
+// +kubebuilder:validation:Enum=kata;enclave-cc
 type CcRuntimeName string
-
-const (
-	// Kata
-	KataCcRuntime CcRuntimeName = "kata"
-
-	// Other Cc Runtime
-)
 
 // CcRuntimeSpec defines the desired state of CcRuntime
 type CcRuntimeSpec struct {
--- a/bundle/manifests/cc-operator.clusterserviceversion.yaml
+++ b/bundle/manifests/cc-operator.clusterserviceversion.yaml
@@ -207,6 +207,13 @@ metadata:
                   }
                 ]
               },
+              "runtimeClassNames": [
+                "kata",
+                "kata-clh",
+                "kata-clh-tdx",
+                "kata-qemu",
+                "kata-qemu-tdx"
+              ],
               "uninstallCmd": [
                 "/opt/kata-artifacts/scripts/kata-deploy.sh",
                 "cleanup"
@@ -350,7 +357,7 @@ spec:
                 - --leader-elect
                 command:
                 - /manager
-                image: controller:latest
+                image: quay.io/confidential-containers/operator:latest
                 livenessProbe:
                   httpGet:
                     path: /healthz
--- a/bundle/manifests/confidentialcontainers.org_ccruntimes.yaml
+++ b/bundle/manifests/confidentialcontainers.org_ccruntimes.yaml
@@ -5555,6 +5555,7 @@ spec:
               runtimeName:
                 enum:
                 - kata
+                - enclave-cc
                 type: string
             required:
             - config
@@ -5632,6 +5633,7 @@ spec:
                 description: Cc Runtime Name
                 enum:
                 - kata
+                - enclave-cc
                 type: string
               totalNodesCount:
                 description: TotalNodesCounts is the total number of worker nodes
--- a/config/manager/kustomization.yaml
+++ b/config/manager/kustomization.yaml
@@ -12,5 +12,5 @@ apiVersion: kustomize.config.k8s.io/v1beta1
 kind: Kustomization
 images:
 - name: controller
-  newName: controller
+  newName: quay.io/confidential-containers/operator
   newTag: latest
--- a/deploy/deploy.yaml
+++ b/deploy/deploy.yaml
@@ -3305,6 +3305,7 @@ spec:
               runtimeName:
                 enum:
                 - kata
+                - enclave-cc
                 type: string
             required:
             - config
@@ -3370,6 +3371,7 @@ spec:
                 description: Cc Runtime Name
                 enum:
                 - kata
+                - enclave-cc
                 type: string
               totalNodesCount:
                 description: TotalNodesCounts is the total number of worker nodes targeted by this CR


jiere:
Agree to add "enclave-cc" here, since all the CoCo sub projects need to share the same general operator.
But how about this file: [deploy.yaml](https://github.com/confidential-containers/operator/blob/main/deploy/deploy.yaml).
In current CoCo operator installation [document](https://github.com/confidential-containers/operator/blob/main/docs/INSTALL.md) uses below command to deploy the operator first:
kubectl apply -f https://raw.githubusercontent.com/confidential-containers/operator/main/deploy/deploy.yaml

I suppose the new runtime enum name("enclave-cc") should take effect there also.

One more concern is on the ccruntime.yaml. Also in the above INSTALL.md, we tell customer to use the operator repo's [ccruntime.yaml](https://github.com/confidential-containers/operator/blob/main/config/samples/ccruntime.yaml) to create the new CR, but currently in this file, the runtime class definition are all Kata specific. So where should enclave-cc' ccruntime.yaml be located?

bpradipt:
> 
> One more concern is on the ccruntime.yaml. Also in the above INSTALL.md, we tell customer to use the operator repo's [ccruntime.yaml](https://github.com/confidential-containers/operator/blob/main/config/samples/ccruntime.yaml) to create the new CR, but currently in this file, the runtime class definition are all Kata specific. So where should enclave-cc' ccruntime.yaml be located?

you can create a new ccruntime yaml eg. ccruntime-enclave-cc.yaml and update the documentation

mythi:
> Agree to add "enclave-cc" here, since all the CoCo sub projects need to share the same general operator. But how about this file: [deploy.yaml](https://github.com/confidential-containers/operator/blob/main/deploy/deploy.yaml). In current CoCo operator installation [document](https://github.com/confidential-containers/operator/blob/main/docs/INSTALL.md) uses below command to deploy the operator first: kubectl apply -f https://raw.githubusercontent.com/confidential-containers/operator/main/deploy/deploy.yaml
> 
> I suppose the new runtime enum name("enclave-cc") should take effect there also.
> 
> One more concern is on the ccruntime.yaml. Also in the above INSTALL.md, we tell customer to use the operator repo's [ccruntime.yaml](https://github.com/confidential-containers/operator/blob/main/config/samples/ccruntime.yaml) to create the new CR, but currently in this file, the runtime class definition are all Kata specific. So where should enclave-cc' ccruntime.yaml be located?

right. `make manifests bundle generate` does not seem to be enough. I need figure our how it's done.

bpradipt:
> > deploy the operator first: kubectl apply -f https://raw.githubusercontent.com/confidential-containers/operator/main/deploy/deploy.yaml
> > I suppose the new runtime enum name("enclave-cc") should take effect there also.
> 
> right. `make manifests bundle generate` does not seem to be enough. I need figure our how it's done.

Something like this
```
diff --git a/api/v1beta1/ccruntime_types.go b/api/v1beta1/ccruntime_types.go
index 811b8c6..27343c9 100644
--- a/api/v1beta1/ccruntime_types.go
+++ b/api/v1beta1/ccruntime_types.go
@@ -24,13 +24,15 @@ import (
 // EDIT THIS FILE!  THIS IS SCAFFOLDING FOR YOU TO OWN!
 // NOTE: json tags are required.  Any new fields you add must have json tags for the fields to be serialized.

-// +kubebuilder:validation:Enum=kata
+// +kubebuilder:validation:Enum=kata;enclave-cc
 type CcRuntimeName string

 const (
        // Kata
        KataCcRuntime CcRuntimeName = "kata"

+       EnclaveCcRuntime CcRuntimeName = "enclave-cc"
+
        // Other Cc Runtime
 )
```

and then `make generate && make bundle` should update the manifests

Makefile
```diff
@@ -160,7 +160,7 @@ ENVTEST ?= $(LOCALBIN)/setup-envtest
 
 ## Tool Versions
 KUSTOMIZE_VERSION ?= v3.8.7
-CONTROLLER_TOOLS_VERSION ?= v0.9.2
+CONTROLLER_TOOLS_VERSION ?= v0.10.0
```
bpradipt, line 163: Please don't change this version. This is closely tied to the sdk version and might have unintended consequence. We can update this once we upgrade the sdk version.

config/crd/bases/confidentialcontainers.org_ccruntimes.yaml
```diff
@@ -3,7 +3,7 @@ apiVersion: apiextensions.k8s.io/v1
 kind: CustomResourceDefinition
 metadata:
   annotations:
-    controller-gen.kubebuilder.io/version: v0.9.2
+    controller-gen.kubebuilder.io/version: v0.10.0
```
bpradipt, line 6: Same as https://github.com/confidential-containers/operator/pull/91/files#r978280082

Makefile
mythi: OK I can drop this.

mythi:
> and then `make generate && make bundle` should update the manifests

@bpradipt It does not seem to touch `deploy.yaml`. Do you want be to keep these unused `const`s?

jiere:
>  you can create a new ccruntime yaml eg. ccruntime-enclave-cc.yaml and update the documentation

Agree, this means we still need to change document in this repo. Also, I am thinking about if this [file](https://github.com/confidential-containers/operator/blob/main/config/samples/kustomization.yaml) change needed?

Makefile
jiere: > Please don't change this version. This is closely tied to the sdk version and might have unintended consequence. We can update this once we upgrade the sdk version.

Which SDK do you mean, is it Operator SDK mentioned [here](https://github.com/confidential-containers/operator/blob/main/docs/DEVELOPMENT.md)? I just want to say that currently in our local setup, the controller side golang code use Kubebuilder annotation to auto-generate stuffs, it looks like we do not have Operator SDK installed but the operator image build is successful.

bpradipt:
> > and then `make generate && make bundle` should update the manifests
> 
> @bpradipt It does not seem to touch `deploy.yaml`. Do you want be to keep these unused `const`s?

Ah yes.. The deploy.yaml was created as a convenience all-in-one single yaml file for deployment using `kubectl` instead of running `make deploy IMG=quay.io/confidential-containers/operator:latest`.
I used the following changes to the Makefile to create the deploy.yaml. May be we can add it as a separate build target.

```
 git diff Makefile
diff --git a/Makefile b/Makefile
index e77154e..3be6e7f 100644
--- a/Makefile
+++ b/Makefile
@@ -140,7 +140,7 @@ uninstall: manifests kustomize ## Uninstall CRDs from the K8s cluster specified
 .PHONY: deploy
 deploy: manifests kustomize ## Deploy controller to the K8s cluster specified in ~/.kube/config.
        cd config/manager && $(KUSTOMIZE) edit set image controller=${IMG}
-       $(KUSTOMIZE) build config/default | kubectl apply -f -
+       $(KUSTOMIZE) build config/default | tee deploy/deploy.yaml | kubectl apply -f -
```

bpradipt:
> Do you want be to keep these unused `const`s?

We don't seem to be using it so it's fine to remove it for now imho.

Makefile
```diff
@@ -88,6 +88,8 @@ help: ## Display this help.
 .PHONY: manifests
 manifests: controller-gen ## Generate WebhookConfiguration, ClusterRole and CustomResourceDefinition objects.
 	$(CONTROLLER_GEN) rbac:roleName=manager-role crd webhook paths="./..." output:crd:artifacts:config=config/crd/bases
+	cd config/manager && $(KUSTOMIZE) edit set image controller=${IMG}
+	$(KUSTOMIZE) build config/default > deploy/deploy.yaml
```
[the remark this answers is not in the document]
bpradipt: Rest of the changes looks good..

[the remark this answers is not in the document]
mythi: I dropped this and updated `deploy/deploy.yaml` manually. Looks like the latest operator-sdk added some CSV changes

bpradipt (approved):
/lgtm
Thanks @mythi

bpradipt:
@arronwy can you please re-run the TDX CI for this PR ?

mythi:
rebased on force pushed

jensfr (approved):
lgtm, thanks!

PR #95, opened 2022-09-27
reconcile: use configurable namespace for the secondary resources

mythi:
CcRuntime is a cluster scoped resource and therefore their instances do not specify any namespace info.

However, the secondary resources are namespaced. Currently, a hard- code value is used but users may want to use some other namespace.

Make the secondary resources namespace configurable using a flag and env var to the operator. By default, use the same (but have it dynamic) namespace as what the operator itself runs in.

Signed-off-by: Mikko Ylinen <mikko.ylinen@intel.com>

### DIFF
--- a/config/manager/manager.yaml
+++ b/config/manager/manager.yaml
@@ -52,5 +52,10 @@ spec:
           requests:
             cpu: 100m
             memory: 20Mi
+        env:
+        - name: CCRUNTIME_NAMESPACE
+          valueFrom:
+            fieldRef:
+              fieldPath: metadata.namespace
       serviceAccountName: controller-manager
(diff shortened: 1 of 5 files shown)


fidencio (approved):
lgtm, thanks @mythi!

fidencio:
I don't recall whether our tests are looking for a specific namespace and whether this will also have to be changed as part of this patch.  Let's run the CI and address those if needed.

bpradipt (approved):
/lgtm
Thanks @mythi 
Looks like this will need a rebase

mythi:
@bpradipt rebased

bpradipt:
@fidencio @wainersm any idea why the CI is stuck ?

PR #99, opened 2022-09-28
ci: Add tdx supported runtime to tests

arronwy:
Both runtime kata-qemu-tdx and kata-clh-tdx verified with operator CI on TDX machine

Signed-off-by: Wang, Arron <arron.wang@intel.com>


wainersm (approved):

bpradipt (approved):
/lgtm

PR #101, opened 2022-09-28
Add ssh-demo sample payload

stevenhorsman:
Add a new payload with the ssh-demo decryption keys and config baked in for demo-ing an encrypted image

Signed-off-by: stevenhorsman <steven@uk.ibm.com>

PR #102, opened 2022-09-29
tests: allow to test alternative ccruntimes

wainersm:
This changes the test scripts so that it is possible to install and run tests for ccruntimes other than the default.

Suppose that you want to test the ccruntime-ssh-demo from PR #101 , just do:

```
$ ./run-local.sh -r kata-qemu -p ccruntime-ssh-demo
```
Ah, and talking about the PR #101, I've converted the instructions @stevenhorsman gave on the PR into a bats test case.

### DIFF
--- a/tests/e2e/ccruntime-ssh-demo_tests.bats
+++ b/tests/e2e/ccruntime-ssh-demo_tests.bats
@@ -1,0 +1,79 @@
+#!/usr/bin/env bats
+# Copyright Confidential Containers Contributors
+#
+# SPDX-License-Identifier: Apache-2.0
+#
+# Implement tests for the ccruntime-ssh-demo runtime.
+#
+load "${BATS_TEST_DIRNAME}/lib.sh"
+test_tag="[cc][ssh demo]"
+
+RUNTIMECLASS="${RUNTIMECLASS:-kata}"
+deployment_file="$BATS_TEST_TMPDIR/ssh-demo.yaml"
+ssh_key_file="$BATS_TEST_TMPDIR/ssh-key"
+
+setup() {
+	cat <<- EOF > "$ssh_key_file"
+	-----BEGIN OPENSSH PRIVATE KEY-----
+	b3BlbnNzaC1rZXktdjEAAAAABG5vbmUAAAAEbm9uZQAAAAAAAAABAAAAMwAAAAtzc2gtZW
+	QyNTUxOQAAACAfiGV2X4o+6AgjVBaY/ZR2UvZp84dVYF5bpNZGMLylQwAAAIhawtHJWsLR
+	yQAAAAtzc2gtZWQyNTUxOQAAACAfiGV2X4o+6AgjVBaY/ZR2UvZp84dVYF5bpNZGMLylQw
+	AAAEAwWYIBvBxQZgk0irFku3Lj1Xbfb8dHtVM/kkz/Uz/l2h+IZXZfij7oCCNUFpj9lHZS
+	9mnzh1VgXluk1kYwvKVDAAAAAAECAwQF
+	-----END OPENSSH PRIVATE KEY-----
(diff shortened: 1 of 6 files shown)


wainersm:
I marked as do-not-merge because I would like to remove some of the TODO's I marked in the code, besides other small changes  like better variable names...etc. The script is functional though.

Let's ensure it doesn't introduce regressions:

/test

wainersm:
The sev job failed because Jenkins was updated today. I need to get https://github.com/kata-containers/ci/pull/499 merged so that the sev job will run on the right node.

wainersm:
Manually changed the sev job on Jenkins. Trying again:

/test-kata-qemu-sev

wainersm:
Made a small fix so that running `./operator.sh install` or `./operator.sh uninstall` doesn't crash.

/test

tests/e2e/ccruntime-ssh-demo_tests.bats
```diff
@@ -0,0 +1,79 @@
+#!/usr/bin/env bats
```
fitzthum, line 1: Seems like maybe this should go in the tests repo? I guess this is fine for the short term.

fitzthum (approved):
I'm kind of on the fence about having multiple payloads. Our end goal is to converge the payload we have as much as possible i.e. one kernel, one rootfs, etc. I think having a debug payload could make sense tho. Maybe in the future we can enable some of the debugging stuff in it.

PR looks fine although it seems like maybe the bats file should be in the tests repo?

Let's just not break anything at the last second.

wainersm:
> I'm kind of on the fence about having multiple payloads. Our end goal is to converge the payload we have as much as possible i.e. one kernel, one rootfs, etc. I think having a debug payload could make sense tho. Maybe in the future we can enable some of the debugging stuff in it.
> 

It is also not clear to me the strategy regarding runtimes on long term. I thought that even with a single ccruntime for CoCo, we might need the tests scripts adapted to enclave-cc's runtime (unless we plan to further merge process-based CoCo runtimes on "regular" VM-based payloads...) 

> PR looks fine although it seems like maybe the bats file should be in the tests repo?
> 

I'm not sure. There is a version of the test in the tests repos already. Sharing tests with kata is cool but IMO we should re-visit that decision or at least re-design the tests as today are hacked to run on operator CI.

> Let's just not break anything at the last second.

Absolutely! I never intended to merge this PR before the release. At this point I consider the CI work for 1st release done; so don't touch it :)

Let's have this PR open for a couple of days in case some else has any comment/feedback. Then we decide whether is worth to merge or not.

fitzthum:
> It is also not clear to me the strategy regarding runtimes on long term. I thought that even with a single ccruntime for CoCo, we might need the tests scripts adapted to enclave-cc's runtime (unless we plan to further merge process-based CoCo runtimes on "regular" VM-based payloads...)

Yeah that's a good point. 

What I like about having a debug payload is that it minimizes the chances of debug features being available in production. Having the keys for an SSH container (with a publicly known SSH key) inside a production pod would be bad, for instance. We could do something similar by adding some debug runtime classes and config files, but we would end up with a lot of runtime classes.

> I'm not sure. There is a version of the test in the tests repos already. Sharing tests with kata is cool but IMO we should re-visit that decision or at least re-design the tests as today are hacked to run on operator CI.

Ok, sounds good. I think we probably don't want to share the non-tee tests in the long run. It will depend a bit on whether we end up switching to kata-deploy for the Kata CI, but I agree that sharing might actually be suboptimal. We are going to be a bit constrained on CI resources with CC HW so we might not want to have the coverage be the same for both projects.

fidencio:
Closing this one as it's stale.

Please, if you plan to rework this, feel free to re-open the PR, rebase this atop of the main, and ping the maintainers of the project for a review.

PR #114, opened 2022-10-07
Import container-engine-for-cc-payload

fidencio:
Let's import all the files from the container-engine-for-cc-payload repo into the Operator one.

The reason for that is for simplifying development and unifying the components used by the Operator.

I've taken the decision to *not* keep the commit history as that was a very short one and not worthwhile, IMHO.

### DIFF
--- a/install/pre-install-payload/Makefile
+++ b/install/pre-install-payload/Makefile
@@ -1,0 +1,23 @@
+CONTAINERD_VERSION=v1.6.6.0
+
+SYSTEMD_DROP_IN_CONFIG = configs/container-engine-for-cc-override.conf.in
+
+OUTPUT_DIR = output
+
+SED = sed
+BASH = bash
+MKDIR = mkdir
+
+containerd: containerd-container-image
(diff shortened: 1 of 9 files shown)


bpradipt (approved):
/lgtm 
Thanks a lot @fidencio

wainersm:
@fidencio great! Small detail: I think you need to add the copyright statement on the files headers.

fidencio:
> @fidencio great! Small detail: I think you need to add the copyright statement on the files headers.

If this is required, we should be testing it to ensure the copyright statement is present on the files, and which files should have them.

wainersm:
> > @fidencio great! Small detail: I think you need to add the copyright statement on the files headers.
> 
> If this is required, we should be testing it to ensure the copyright statement is present on the files, and which files should have them.

I don't know if CNCF enforces that. If yes, then we will need to indeed add a checker for each and every PR.

We can merge this PR as is and address this if needed in a follow up PR.

PR #119, opened 2022-11-02
deploy: add CCRUNTIME_NAMESPACE env to operator

mythi:
This should take `deploy.yaml` up-to-date with `manager.yaml` to be able to use `:latest` image.

### DIFF
--- a/deploy/deploy.yaml
+++ b/deploy/deploy.yaml
@@ -3710,6 +3710,11 @@ spec:
           requests:
             cpu: 100m
             memory: 20Mi
+        env:
+        - name: CCRUNTIME_NAMESPACE
+          valueFrom:
+            fieldRef:
+              fieldPath: metadata.namespace
         securityContext:
           allowPrivilegeEscalation: false
       securityContext:


bpradipt:
@mythi based on your earlier suggestion to use kustomize (`kubectl apply -k ..`) and doing away with deploy.yaml,  do we really want this change ?

mythi:
> @mythi based on your earlier suggestion to use kustomize (`kubectl apply -k ..`) and doing away with deploy.yaml, do we really want this change ?

I'm fine dropping this. This is the fix for `deploy.yaml` if we want to keep it for `v0.2.0` still.

bpradipt:
> > @mythi based on your earlier suggestion to use kustomize (`kubectl apply -k ..`) and doing away with deploy.yaml, do we really want this change ?
> 
> I'm fine dropping this. This is the fix for `deploy.yaml` if we want to keep it for `v0.2.0` still.

I would prefer dropping deploy.yaml and changing the docs to use the kustomize way. The earlier the better imho.
But let's hear from few more folks and we can decide..  

@fidencio @wainersm @jensfr @fitzthum please comment..

fidencio:
> > > @mythi based on your earlier suggestion to use kustomize (`kubectl apply -k ..`) and doing away with deploy.yaml, do we really want this change ?
> > 
> > 
> > I'm fine dropping this. This is the fix for `deploy.yaml` if we want to keep it for `v0.2.0` still.
> 
> I would prefer dropping deploy.yaml and changing the docs to use the kustomize way. The earlier the better imho. But let's hear from few more folks and we can decide..
> 
> @fidencio @wainersm @jensfr @fitzthum please comment..

@mythi, I'm with @bpradipt on this one.
Would it be possible to close this one and open a new one updated the docs to use the kustomize way?

mythi:
I'm OK closing.

PR #129, opened 2022-11-11
config: allow generation of bundle from scratch

jensfr:
We don't need the generated bundle files in the repository. It can be generated via 'make bundle'. Any changes to manifests etc. should be made in the correct places under config/* or via kubebuilder annotations in the controller code.

The diff between files generated via 'make bundle' and what we currently have is minimal and actually desired:

image

Let's go through the changes one by one,

1. runtime class names, the same classes but different order

2. a new version of kube-rbac-proxy which may be caused by a new controller-gen version

3. CCRUNTIME_NAMESPACE is used in main.go and was added by

  09e9b8c6 (Mikko Ylinen  2022-09-27 12:18:30 +0300 56)

Given the above, I think it is safe to remove the bundle files from the repository.

Signed-off-by: Jens Freimann <jfreimann@redhat.com>


mythi:
> Given the above, I think it is safe to remove the bundle files from the repository.

+1. I had this in my mind earlier as well but looks like I forgot to add an issue about it. 

>  2. a new version of kube-rbac-proxy which may be caused by a new controller-gen version

controller-gen does not update it. The new image has a bunch of CVEs fixed so it was updated manually.

fidencio (approved):
lgtm, thanks @jensfr!

config/samples/ccruntime.yaml
```diff
@@ -2,6 +2,7 @@ apiVersion: confidentialcontainers.org/v1beta1
 kind: CcRuntime
 metadata:
   name: ccruntime-sample
+  namespace: confidential-containers-system
```
mythi, line 5: is there a need to add this back?

jensfr: I'll double check and update. Thanks for the reviews!

fidencio:
Closing this one as it's stale.

Please, if you plan to rework this, feel free to re-open the PR, rebase this atop of the main, and ping the maintainers of the project for a review.

PR #133, opened 2022-11-21
actions: Add a proto CI/CD for Enclave CC

fidencio:
Let's add the bare minimum CI / CD for Enclave CC, running a GitHub action once a day to make sure the latest published CI payload works as expected.

### DIFF
--- a/.github/workflows/enclave-cc-cicd.yaml
+++ b/.github/workflows/enclave-cc-cicd.yaml
@@ -1,0 +1,59 @@
+name: CI/CD | enclave-cc e2e tests
+on:
+  schedule:
+  - cron: "0 0 * * *"
+
+permissions:
+  contents: read
+
+jobs:
+  e2e:
+    name: operator enclave-cc e2e tests
+    runs-on: ubuntu-22.04
+    services:
+      registry:
+        image: registry:2
+        ports:
+          - 5000:5000
+    steps:
+      - uses: actions/checkout@v3
+
+      - name: Set up Docker Buildx
+        uses: docker/setup-buildx-action@v2
+        with:
+          driver-opts: network=host
+
+      - name: Build and push to local registry
+        run: |
+          make docker-build
+          make docker-push
+        env:
+          IMG: localhost:5000/cc-operator:latest
+
+      - name: Setup kind cluster
+        run: |
+          kind create cluster --image "kindest/node:v1.24.0" -n coco-sgx --config tests/e2e/enclave-cc-kind-config.yaml --wait 120s
+          kubectl label node coco-sgx-worker node-role.kubernetes.io/worker=
+
+      - name: Deploy operator from the local registry
+        run: |
+          cd config/default
+          kustomize edit set image localhost:5000/cc-operator:latest
+          kubectl apply -k .
+          sleep 1
+          kubectl wait --for=jsonpath='{.status.conditions[0].status}'=True deployments/cc-operator-controller-manager -n confidential-containers-system
+
+      - name: Install enclave-cc sim
+        run: |
+          current_tag=$(curl -X GET "https://quay.io/api/v1/repository/confidential-containers/runtime-payload/tag/" | jq .tags[].name | grep enclave-cc-SIM | grep -v latest | head -1 | sed -e 's|\"||g')
+          latest_ci_tag=$(curl -X GET "https://quay.io/api/v1/repository/confidential-containers/runtime-payload-ci/tag/" | jq .tags[].name | grep enclave-cc-SIM | grep -v latest | head -1 | sed -e 's|\"||g') 
+          sed -e -i 's|runtime-payload:${current_tag}|runtime-payload-ci:${latest_ci_tag}/g' config/samples/ccruntime-enclave-cc-sim.yaml
+          echo "Testing Enclave CC using the following CI tag: ${latest_ci_tag}"
+          kubectl apply -f config/samples/ccruntime-enclave-cc-sim.yaml
+          sleep 1
+          kubectl wait --for=jsonpath='{.status.runtimeClass}'=enclave-cc ccruntime/ccruntime-enclave-cc-sim --timeout=90s
+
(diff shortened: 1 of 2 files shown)


.github/workflows/enclave-cc-cicd.yaml
```diff
@@ -0,0 +1,59 @@
...
+          kubectl apply -k .
+          sleep 1
+          kubectl wait --for=jsonpath='{.status.conditions[0].status}'=True deployments/cc-operator-controller-manager -n confidential-containers-system
+
+      - name: Install enclave-cc sim
+        run: |
+          current_tag=$(curl -X GET "https://quay.io/api/v1/repository/confidential-containers/runtime-payload/tag/" | jq .tags[].name | grep enclave-cc-SIM | grep -v latest | head -1 | sed -e 's|\"||g')
+          latest_ci_tag=$(curl -X GET "https://quay.io/api/v1/repository/confidential-containers/runtime-payload-ci/tag/" | jq .tags[].name | grep enclave-cc-SIM | grep -v latest | head -1 | sed -e 's|\"||g') 
+          sed -e -i 's|runtime-payload:${current_tag}|runtime-payload-ci:${latest_ci_tag}/g' config/samples/ccruntime-enclave-cc-sim.yaml
+          echo "Testing Enclave CC using the following CI tag: ${latest_ci_tag}"
```
fidencio, line 51: @mythi, could I use `kustomize` to do such change?

.github/workflows/enclave-cc-cicd.yaml
```diff
@@ -0,0 +1,59 @@
...
+      - name: Build and push to local registry
+        uses: docker/build-push-action@v3
+        with:
+          context: .
+          push: true
+          tags: localhost:5000/cc-operator:latest
+
+      - name: Setup kind cluster
+        run: |
+          kind create cluster -n coco-sgx --config tests/e2e/enclave-cc-kind-config.yaml
```
wainersm, line 35: is `kind` installed in the environment?

fidencio: Yes, it's.  I've copied this from the https://github.com/confidential-containers/operator/blob/main/.github/workflows/e2e.yaml, which is up and running with this as it is. :-)

.github/workflows/enclave-cc-cicd.yaml
```diff
@@ -0,0 +1,59 @@
...
+          - 5000:5000
+    steps:
+      - uses: actions/checkout@v3
+
+      - name: Set up Docker Buildx
+        uses: docker/setup-buildx-action@v2
+        with:
+          driver-opts: network=host
+
+      - name: Build and push to local registry
```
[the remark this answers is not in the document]
fidencio: My understanding is that this is exactly what this action does.  Plus, it aligns with https://github.com/confidential-containers/operator/blob/main/.github/workflows/e2e.yaml

.github/workflows/enclave-cc-cicd.yaml
mythi: yes `kind` is pre-installed in Github virtual envs.

wainersm: You also could pass: 
 * `--wait duration` so it wait the control node be ready.
 * --image "kindest/node:v1.24.0" to use k8s 1.24

.github/workflows/enclave-cc-cicd.yaml
```diff
@@ -0,0 +1,59 @@
...
+  - cron: "0 0 * * *"
+
+permissions:
+  contents: read
+
+jobs:
+  e2e:
+    name: operator enclave-cc e2e tests
+    runs-on: ubuntu-22.04
+    services:
```
wainersm, line 13: When I created the e2e tests for the operator I used an external registry but It seems I didn't need that.

On confidential-containers/cloud-api-adaptor, there is a CI for a webhook which leverages `Kind` as well. There, we are pushing the controller image to the cluster with `kind load` command (see: https://kind.sigs.k8s.io/docs/user/quick-start/#loading-an-image-into-your-cluster). There is one details that made me getting stalled until @bpradipt came to rescue me: the controller image *must* be tagged and the tag *must not* be *latest*.

mythi: right. `kind load` makes the image available but the `ImagePullPolicy` overrides that local image. That's why a local registry works nicely.

.github/workflows/enclave-cc-cicd.yaml
mythi: we're missing `kustomization`s for enclave-cc `CcRuntime` but I can see that could indeed help: it would be possible to set `payloadImage` easily to any (sim/hw/CI/published) images

.github/workflows/enclave-cc-cicd.yaml
wainersm: > Yes, it's. I've copied this from the https://github.com/confidential-containers/operator/blob/main/.github/workflows/e2e.yaml, which is up and running with this as it is. :-)

oh, just realized this PR is a copy of the an existing e2e workflow.

.github/workflows/enclave-cc-cicd.yaml
wainersm: yep, the image must be tagged and not `latest` so that the `IfNotPresent` is set by default, as explained in https://kind.sigs.k8s.io/docs/user/quick-start/#loading-an-image-into-your-cluster

.github/workflows/enclave-cc-cicd.yaml
fidencio: Yep, basically that, but with a few change to actually use the payload generated for the Ci, and to run it every day.

I've propagated the changes you suggested to that workflow as well.

wainersm:
I keep my approval as @fidencio only fixed the --wait value being passed to kind.

bpradipt (approved):
/lgtm
thanks @fidencio

PR #134, opened 2022-11-21
actions: Ensure payloads are using the latest tag

fidencio:
Let's ensure the payloads are always using the latest released tag.


.github/workflows/runtime-payload-latest-tags.yaml
```diff
@@ -0,0 +1,62 @@
+name: Ensure the latest tags are used for the runtime payloads
```
wainersm, line 1: I talked with @fidencio on private about this PR a little bit. My suggestion is to make it part of a release script. Below is my rationale:

* in general I don't like the idea of run actions to catch problems not related with the proposed changes of a PR. The reasons being that 1) it can fail-block the PR and it's not the owners changes fault, 2) it pollute the checks reports, 3) there isn't any guarantee someone will fix the errors caught   
* IIUC the payloads tag will be updated only and only if there is a release. There is not point in keep it running for 6 weeks (current release cadence) when we need it only at release time.

wainersm: @fitzthum is there any plan to have a script(s) to automate the release?

fidencio: I agree on having these checks as part of some release script, that's way better!

/cc @bpradipt and @jensfr as well.

fitzthum: Yeah I agree with everything here. We might also want to think about the freeze window vs the rest of development in terms of what kind of checks we want.

fidencio:
I'm closing this one  :-)

PR #136, opened 2022-11-23
add kustomization for enclave-cc deployments

mythi:
in an ealier PR @fidencio  [was asking](https://github.com/confidential-containers/operator/pull/133#discussion_r1027803323) can we simplify things using `kustomize`. Yes we can. With this change, the maintenance of `CcRuntime` customizations for enclave-cc is made simple. Perhaps others can look into it too?


fidencio (approved):
lgtm, thanks @mythi!

jensfr (approved):
lgtm! Thanks @mythi ! We should use kustomize where possible.

PR #143, opened 2022-12-19
ccruntime: Test with the latest kata payload

fidencio:

### DIFF
--- a/config/samples/ccruntime/default/kustomization.yaml
+++ b/config/samples/ccruntime/default/kustomization.yaml
@@ -10,12 +10,13 @@ images:
 - name: quay.io/confidential-containers/container-engine-for-cc-payload
   newTag: 1034f9fcf947b22eea080a6f77d8e164e2369849
 - name: quay.io/confidential-containers/runtime-payload
-  newTag: kata-containers-5b7009f2f9cdb39099b97a4f868b53ec9f3c383f
+  newName: quay.io/confidential-containers/runtime-payload-ci
+  newTag: kata-containers-latest
 
 patches:
 - patch: |-
     - op: replace
       path: /spec/config/runtimeClassNames
-      value: ["kata", "kata-clh", "kata-clh-tdx", "kata-clh-tdx-eaa-kbc", "kata-qemu", "kata-qemu-tdx", "kata-qemu-tdx-eaa-kbc", "kata-qemu-sev"]
+      value: ["kata", "kata-clh", "kata-clh-tdx", "kata-qemu", "kata-qemu-tdx", "kata-qemu-sev"]
   target:
     kind: CcRuntime
--- a/tests/e2e/tests_runner.sh
+++ b/tests/e2e/tests_runner.sh
@@ -70,6 +70,7 @@ usage() {
 # tests for CC without specific hardware support
 run_non_tee_tests() {
 	local runtimeclass="$1"
+	local aa_kbc="${2:-"offline_fs_kbc"}"
 
 	# Results for agent_image.bats:
 	#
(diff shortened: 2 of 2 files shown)


fidencio:
@ryansavino, @UnmeshDeodhar, mind to take a look on what's going on with the SEV machine?

fidencio:
Okay, with this changes we have the current `:latest` payload from the CI repo (which builds a new payload on every merge that happens in the `CCv0` branch of Kata Containers) passing the tests.

fidencio:
> @ryansavino, @UnmeshDeodhar, mind to take a look on what's going on with the SEV machine?

Now it seems like a legit failure, please, take a look.

ryansavino:
Nothing jumping out at me from the logs initially. Let me take a further look today and see if I can figure out what's going on.

fidencio:
> Nothing jumping out at me from the logs initially. Let me take a further look today and see if I can figure out what's going on.

```
⋊> ffidenci at ffidenci-mobl1 ⋊> tests on CCv0 ≡ ⨯ git diff
diff --git a/integration/kubernetes/confidential/sev.bats b/integration/kubernetes/confidential/sev.bats
index d532fbf38..cef4c2709 100644
--- a/integration/kubernetes/confidential/sev.bats
+++ b/integration/kubernetes/confidential/sev.bats
@@ -178,7 +178,7 @@ generate_firmware_measurement_with_append() {
 local append_default="tsc=reliable no_timer_check rcupdate.rcu_expedited=1 \
 i8042.direct=1 i8042.dumbkbd=1 i8042.nopnp=1 i8042.noaux=1 noreplace-smp reboot=k \
 cryptomgr.notests net.ifnames=0 pci=lastbus=0 console=hvc0 console=hvc1 quiet panic=1 \
-nr_cpus=1 scsi_mod.scan=none agent.config_file=/etc/agent-config.toml"
+nr_cpus=1 selinux=0 scsi_mod.scan=none agent.config_file=/etc/agent-config.toml"
 
   local sev_config="/opt/confidential-containers/share/defaults/kata-containers/configuration-qemu-sev.toml"
 ```

This should solve the issue.

**Comment by fitzthum** [comment]
So this is not related to the SELinux thing. The line that @fidencio mentions above is just the default kernel command line, which gets overwritten by the command line that was actually used in the previous run. The true command line is printed in the log (right above the test failure message) and you can see that the selinux param is included. 

Interestingly the test that is failing here is actually the negative test where we try to start a container image with an invalid launch measurement and expect it to fail. In this case the container actually starts, which is not ideal. 

The issue comes down to these lines in the test file 
```
INSERT INTO secrets VALUES (10, 'key_id1', '${ENCRYPTION_KEY}', 10);
INSERT INTO keysets VALUES (10, 'KEYSET-1', '["key_id1"]', 10);
INSERT INTO policy VALUES (10, '["${measurement}"]', '[]', 0, 0, '[]', now(), NULL, 1);
```
This is where we provision the KBS with the expected measurement by setting a new policy. We then assign this policy both to `key_id1` and `KEYSET-1`. For the offline KBC this is actually redundant because `key_id1` is part of `KEYSET-1` so the policy is evaluated twice. Not a big deal.

With the online KBS that we just switched to things are a bit different. The online KBS does not use keysets (technically a user could request a keyset at runtime, but we aren't doing that in this test). Instead we request a connection at launch. The connection will be granted as long as the guest meets the KBS-wide default policy. The idea is that policy evaluation should mainly happen at runtime when individual resources/keys are requested. Since we set a policy for `key_id1` this should be enough to prevent the invalid guest from running the container. Unfortunately here we hit the known issue with the encrypted test image, which was erroneously overwritten with an unencrypted image. With the offline KBS this test still failed (as expected) because the keyset is requested ahead of time, but when we switched to the online KBS the connection request is more permissive and we never actually request `key_id1` with an unencrypted image. 

The good news is that @ryansavino and I just (hopefully) fixed the test image, so this test should go back to breaking. The only thing to keep in mind is that previously the KBS would report an error at boot. Now we will get an error at runtime when we try to decrypt the image. The error message that we are looking for in the test might change so we might have to make a minor adjustment. Hopefully that makes sense.

bpradipt (approved):
/lgtm
Thanks @fidencio

stevenhorsman (approved):
LGTM. Thanks @fidencio

PR #150, opened 2023-01-16
config: Update preInstall payload

fidencio:
Let's point to the payload generated from the 1def978fdf commit, which includes the bump to containerd v1.6.8.1, and has a multi-arch image.


bpradipt (approved):
/lgtm

Issue #168, opened 2023-03-06
Update payloads to latest post 0.4 release


stevenhorsman:
We've had some issues in the past two releases where we've not done a great job at testing the latest versions of the various components as we go along. Now the CoCo 0.4 operator has been tagged, we could switch the payloads to point to the latest ci builds and that would at least let us re-run the operator CI tests throughout the release to check that the latest kata artifacts are working ok.

I'll admit it isn't a perfect solution as we'll still need to remember to run the tests and bump further upstream dependencies (e.g. attestation-agent, image-rs), but I think it is better than not testing the operator until the release process has started.

PR #169, opened 2023-03-06
config: Update payloads to the latest CI image

stevenhorsman:
- Now 0.4 release has been tagged, update the config to point to the latest runtime-payloads for 'CI' testing purposes

Fixes: #168

### DIFF
--- a/config/samples/ccruntime/default/kustomization.yaml
+++ b/config/samples/ccruntime/default/kustomization.yaml
@@ -8,10 +8,10 @@ resources:
 
 images:
 - name: quay.io/confidential-containers/container-engine-for-cc-payload
-  newTag: 98a790e8abdcc06c4b629b290ebaa217bf82e305
+  newTag: latest
 - name: quay.io/confidential-containers/runtime-payload
-  newName: quay.io/confidential-containers/runtime-payload
-  newTag: kata-containers-6abfb9deadc03154f91896e78ea024e55928c9f4-x86_64
+  newName: quay.io/confidential-containers/runtime-payload-ci
+  newTag: kata-containers-latest
 
 patches:
 - patch: |-
--- a/config/samples/ccruntime/peer-pods/kustomization.yaml
+++ b/config/samples/ccruntime/peer-pods/kustomization.yaml
@@ -8,10 +8,10 @@ resources:
 
 images:
 - name: quay.io/confidential-containers/container-engine-for-cc-payload
-  newTag: 98a790e8abdcc06c4b629b290ebaa217bf82e305
+  newTag: latest
 - name: quay.io/confidential-containers/runtime-payload
-  newName: quay.io/confidential-containers/runtime-payload
-  newTag: kata-containers-6abfb9deadc03154f91896e78ea024e55928c9f4-x86_64
+  newName: quay.io/confidential-containers/runtime-payload-ci
+  newTag: kata-containers-latest
 
 patches:
 - patch: |-
--- a/config/samples/ccruntime/s390x/kustomization.yaml
+++ b/config/samples/ccruntime/s390x/kustomization.yaml
@@ -8,10 +8,10 @@ resources:
 
 images:
 - name: quay.io/confidential-containers/runtime-payload
-  newName: quay.io/confidential-containers/runtime-payload
-  newTag: kata-containers-6abfb9deadc03154f91896e78ea024e55928c9f4-s390x
+  newName: quay.io/confidential-containers/runtime-payload-ci
+  newTag: kata-containers-latest
 - name: quay.io/confidential-containers/container-engine-for-cc-payload
-  newTag: 98a790e8abdcc06c4b629b290ebaa217bf82e305
+  newTag: latest
 
 patches:
 - patch: |-
--- a/config/samples/ccruntime/s390x/kustomization.yaml
+++ b/config/samples/ccruntime/s390x/kustomization.yaml
@@ -4,14 +4,7 @@ kind: Kustomization
 nameSuffix: -s390x
 
 resources:
-- ../base
-
-images:
-- name: quay.io/confidential-containers/runtime-payload
-  newName: quay.io/confidential-containers/runtime-payload-ci
-  newTag: kata-containers-latest
-- name: quay.io/confidential-containers/container-engine-for-cc-payload
-  newTag: latest
+- ../default
 
 patches:
 - patch: |-
--- a/config/samples/enclave-cc/base/ccruntime-enclave-cc.yaml
+++ b/config/samples/enclave-cc/base/ccruntime-enclave-cc.yaml
@@ -9,7 +9,7 @@ spec:
       node-role.kubernetes.io/worker: ""
   config:
     installType: bundle
-    payloadImage: quay.io/confidential-containers/runtime-payload:enclave-cc-HW-v0.5.0
+    payloadImage: quay.io/confidential-containers/runtime-payload-ci:enclave-cc-HW-latest
     installDoneLabel:
       confidentialcontainers.org/enclave-cc: "true"
     uninstallDoneLabel:
--- a/config/samples/enclave-cc/sim/kustomization.yaml
+++ b/config/samples/enclave-cc/sim/kustomization.yaml
@@ -4,5 +4,5 @@ resources:
 nameSuffix: -sgx-mode-sim
 
 images:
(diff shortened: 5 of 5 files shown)


fidencio:
@stevenhorsman, this is orthogonal to this PR, but what about adding a nightly job just to make sure that every single day we test those changes?

This is what we've done for Enclave CC and I do believe all payloads would benefit of this approach.

stevenhorsman:
> @stevenhorsman, this is orthogonal to this PR, but what about adding a nightly job just to make sure that every single day we test those changes?
> 
> This is what we've done for Enclave CC and I do believe all payloads would benefit of this approach.

Yeah, that sounds like a good idea, as long as we aren't stressing the kata-containers CI too much

fidencio (approved):
lgtm, thanks @stevenhorsman!

wainersm:
Hi @stevenhorsman @fidencio ,

The problem of using kata-containers-latest is that a break on operator CI doesn't mean it is caused by a change on the operator itself.

> @stevenhorsman, this is orthogonal to this PR, but what about adding a nightly job just to make sure that every single day we test those changes?
> 

I've a PoC of a job in http://jenkins.katacontainers.io/blue/organizations/jenkins/confidential-containers-ci/activity that is triggered for any change on kata-containers, kata-containers/tests or confidential-containers/operator repositories. The job waits the runtime-payload-ci image to show up then trigger children jobs that are going to leverage the new payload built. It has potential to be a true CI for confidential containers.

Thus, I suggest to drop this PR so we keep the operator CI running on a stable payload.

> This is what we've done for Enclave CC and I do believe all payloads would benefit of this approach.

stevenhorsman:
> Thus, I suggest to drop this PR so we keep the operator CI running on a stable payload.

Just to confirm I understand - your CI approach (which sounds awesome), should automatically test the `runtime-payload-ci:latest` with the operator, so you are suggesting this PR be closed without merging so that we only bump the runtime-payload in the operator during the release process?

I guess my initial reaction to this (if I'm being greedy) is that if we are testing the latest runtime in the operator, then surely anyone using `operator main`/`operator:latest` container image should have access to these tested changes as well and not be running backlevel? I guess it depends on how often we see failures and can't fix them, but I think the aim should be that if the 'CI' test fails (and therefore using the `runtime-payload-ci:latest` is a 'bad' thing and unstable) then it should be top priority to get that fixed ASAP, which might mean reverting changes in kata-cc side if the fix isn't a simple one. Maybe that is too idealistic though!

stevenhorsman:
> Okay, IIUC, to be practical we have two options:
> 
> a) Use runtime-payload-ci on a known and tested version in both kata-cc CI and CC CI, but it might require occasional bumps manually so that developers pick up the latest changes. The payload here is always tested on both CIs. b) Use runtime-payload-ci on `latest` and assume it is stable because it passed the kata-cc CI. In case the CC CI detects a failure then we do our best to fix the problem, in that time frame `runtime-payload-ci:latest` can be considered unstable
> 
> Did I understand correctly?
> 
> Any missing case? I'm wondering if a decision here can somehow be bad/good for peer pods too.

Can we both have our cake and eat it I wonder here: It's more effort, so maybe something for the future, but could we have our 'CC CI' job that tests the operator behaviour for a particular version of the `runtime-payload-ci:kata-containers-<sha-1>-<arch>` and then if that is successful push it to a tag like `latest-tested` or something like that then between releases we could have the operator point to this tag and therefore we should always be on a 'stable'(or at least tested) runtime?

wainersm:
> few notes:
> 
> first, I think it's a good idea to create a separate job for testing the latest bundles. It doesn't really make sense to wait for someone to open a PR on the operator to test the latest bundle. Furthermore when someone opens a PR on the operator, it probably is best to isolate that change from everything that is happening upstream for the next release.
> 
> That said, I only heard about this new job because Steve pointed me to this issue. If this is going to be the main green light for the project, can we document and expose it a little bit more clearly?
> 

Sure. I started working on that job (being called "CC CI") on Dec of last year. I finally found some time to up it running two weeks ago (or so). I didn't broadly announced it because I am observing if it is stable (e.g. I had to apply a fix already https://github.com/kata-containers/ci/pull/540 related with stability) and there are still some features I would like to implement (e.g. do not re-build the operator, instead use the one already built).

Give me a few days/weeks and I will have it in a better shape for announcing.
   
> I think we might also be overlooking one usecase of the operator CI; local development. in my experience it is much easier to use `run-local.sh`than to try to replicate the Kata CI locally. This is how I test any changes to the tests, how I install my own custom bundles, and how I setup systems with the latest upstream bundles. Obviously these changes make this a bit trickier (although it's trivial to switch to another bundle if you know how to use Kustomize). I think it might be handy to create a new `latest` overlay via Kustomize that would point to the upstream bundles. We could also add a `latest` flag to `run-local.sh` to switch to this. I think this would make things a little bit more clear. I can implement if people think this is good idea.

I think it is a good idea! :)

Indeed, I personally dislike any CI that you can't easily run locally.

wainersm (approved):

ryansavino:
@wainersm will this allow operator to be run against latest `kata-containers` (CCv0) and `tests` (CCv0)? If not integrated directly with CI for now due to pending merge to main stuff, it may be useful for now to have it as a separate yaml config so that we can run it locally. Last release (.4) can get a bit stale regarding recent changes, and it would be nice to be able to run the operator against more recent, latest changes without having to manually change a bunch of things. What do you think?

wainersm:
> @wainersm will this allow operator to be run against latest `kata-containers` (CCv0) and `tests` (CCv0)? If not integrated directly with CI for now due to pending merge to main stuff, it may be useful to have it as a separate yaml config so that we can run it locally. Last release (.4) can get a bit old, and it would be nice to be able to run the operator against more recent, latest changes without having to manually change a bunch of things. What do you think?

I agree @ryansavino , let's try to get it merged asap.

The jobs running nightly are mostly broken (http://jenkins.katacontainers.io/view/Daily%20CCv0%20baseline/) because the tests code moved forward while the kata-containers payload is still old.

stevenhorsman:
@wainersm - we've got all the test passing now, so I don't know if this PR is worth merging to test the latest code before we cut it, or if you just want to go straight to using tagged versions?

wainersm:
> @wainersm - we've got all the test passing now, so I don't know if this PR is worth merging to test the latest code before we cut it, or if you just want to go straight to using tagged versions?

@larrydewey should be about to open a PR to test the latest CI payload so it is not worth merging at this point. But we could leave it open and merge after the release (this way we keep the discussion history).

wainersm:
Let's start testing with latest payload. Better later than never :)

/test

stevenhorsman:
> Let's start testing with latest payload. Better later than never :)
> 
> /test

FYI - this might fail now as I updated it to support multi-arch payload images, which I don't think has successfully run yet (Fabiano's working on the fix now for it). If we don't have success I can drop the second commit

stevenhorsman:
> > Let's start testing with latest payload. Better later than never :)
> > /test
> 
> FYI - this might fail now as I updated it to support multi-arch payload images, which I don't think has successfully run yet (Fabiano's working on the fix now for it). If we don't have success I can drop the second commit

The multi-arch payloads has built now, so this should be good (though the tests seemed to pass anyway?)
/test

stevenhorsman:
> The multi-arch payloads has built now, so this should be good (though the tests seemed to pass anyway?)

Okay I think I understand how it passed - Choi re-queued the failing s390x test and by that I managed to get the multi-arch payload built. Thanks Choi!

arronwy:
> @arronwy - would you be able to check if the tdx job that is failing is a real error still? Thanks!

Hi @stevenhorsman

TDX CI failed with
```
13:21:31 # sed: can't read : No such file or directory
13:21:31 # /opt/confidential-containers/share/defaults/kata-containers/configuration-qemu-tdx.toml: file /opt/confidential-containers/bin/qemu-system-x86_64-tdx-experimental does not exist
```

Should be fixed with this PR :https://github.com/kata-containers/kata-containers/pull/6919 maybe we need use new payload image with this fix?

stevenhorsman:
@fidencio @wainersm - can you re-review as I have updated to add mutli-arch support here that you haven't seen, so it would be a bit naughty to merge based on your old reviews!

fidencio (approved):
lgtm, thanks!

wainersm (approved):
hey @stevenhorsman , LGTM. Sorry for blocking this PR for many occasions.

Issue #173, opened 2023-03-09
operator latest image not pushed to quay.io


stevenhorsman:
**Describe the bug**
In PR https://github.com/confidential-containers/operator/pull/165 I attempted to add a workflow to ensure that the `operator:latest` is rebuilt and updated on pushing to `main`, however now it has been merged and https://github.com/confidential-containers/operator/pull/171 was merged we can see from the [registry](https://quay.io/repository/confidential-containers/operator?tab=history) that although the [workflow ran successfully](https://github.com/confidential-containers/operator/actions/runs/4354326095) the push didn't occur.

I need to debug this once I have time and try and work out what is wrong - probably in my own fork and quay repository.

PR #174, opened 2023-03-09
workflow: Add push field to action

stevenhorsman:
Add `push: true` to the build and push action,
so it actually pushes the container image!

Fixes: #173

### DIFF
--- a/.github/workflows/docker-publish-latest-on-merge.yml
+++ b/.github/workflows/docker-publish-latest-on-merge.yml
@@ -64,6 +64,7 @@ jobs:
         uses: docker/build-push-action@v3
         with:
           context: .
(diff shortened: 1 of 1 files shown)


stevenhorsman:
Now I'm doing the fix when not rushing, I've tested it on my fork:
GHA https://github.com/stevenhorsman/cc-operator/actions/runs/4376953421 produced the `main` and `latest` tagged images in https://quay.io/repository/stevenhorsman/cc-operator?tab=tags

fidencio (approved):
lgtm, thanks @stevenhorsman!

wainersm (approved):
It is indeed weird that the build-push-action has push disabled by default...

PR #180, opened 2023-03-14
Update payloads to using the ubuntu base image

fidencio:
---

enclave-cc: Fix preInstall image tag / hostPath / mountPath

We've updated the preInstall image to use Ubuntu 20.04 instead of
CentOS, but we didn't fully reflect the changes for Enclave CC.

---

payloads: Update to an image using ubuntu as base

Both Enclave CC and Kata Containers have switched to using a base image
for their payloads, and we should make sure it works on our side.

As part of the change, we need to update the hostPath / mountPath to
reflect what's needed in order to be able to call `systemctl restart
...` using the new Ubuntu base image.

---

### DIFF
--- a/config/samples/ccruntime/default/kustomization.yaml
+++ b/config/samples/ccruntime/default/kustomization.yaml
@@ -10,8 +10,8 @@ images:
 - name: quay.io/confidential-containers/container-engine-for-cc-payload
   newTag: 98a790e8abdcc06c4b629b290ebaa217bf82e305
 - name: quay.io/confidential-containers/runtime-payload
-  newName: quay.io/confidential-containers/runtime-payload
-  newTag: kata-containers-129e43d1ea5cca528b7b97234b7561219208a244-x86_64
+  newName: quay.io/confidential-containers/runtime-payload-ci
+  newTag: kata-containers-121892ec61cb7d50c662133bb6fbf25b50ab676e-x86_64
 
 patches:
 - patch: |-
--- a/config/samples/enclave-cc/base/ccruntime-enclave-cc.yaml
+++ b/config/samples/enclave-cc/base/ccruntime-enclave-cc.yaml
@@ -9,7 +9,7 @@ spec:
       node-role.kubernetes.io/worker: ""
   config:
     installType: bundle
-    payloadImage: quay.io/confidential-containers/runtime-payload:enclave-cc-HW-v0.4.0
+    payloadImage: quay.io/confidential-containers/runtime-payload-ci:enclave-cc-HW-4370700439f5524a60cf7a31c7c4f92bbe4dd655
     installDoneLabel:
       confidentialcontainers.org/enclave-cc: "true"
     uninstallDoneLabel:
@@ -21,9 +21,9 @@ spec:
         name: enclave-cc-conf
       - mountPath: /opt/confidential-containers/
         name: enclave-cc-artifacts
-      - mountPath: /var/run/dbus
+      - mountPath: /var/run/dbus/system_bus_socket
         name: dbus
-      - mountPath: /run/systemd
+      - mountPath: /run/systemd/system
         name: systemd
       - mountPath: /usr/local/bin/
         name: local-bin
@@ -41,11 +41,11 @@ spec:
           type: DirectoryOrCreate
         name: enclave-cc-artifacts
       - hostPath:
-          path: /var/run/dbus
+          path: /var/run/dbus/system_bus_socket
           type: ""
         name: dbus
       - hostPath:
-          path: /run/systemd
+          path: /run/systemd/system
           type: ""
         name: systemd
       - hostPath:
@@ -63,9 +63,9 @@ spec:
           name: confidential-containers-artifacts
         - mountPath: /etc/systemd/system/
           name: etc-systemd-system
-        - mountPath: /var/run/dbus
+        - mountPath: /var/run/dbus/system_bus_socket
           name: dbus
-        - mountPath: /run/systemd
+        - mountPath: /run/systemd/system
           name: systemd
       volumes:
         - hostPath:
--- a/config/samples/enclave-cc/base/ccruntime-enclave-cc.yaml
+++ b/config/samples/enclave-cc/base/ccruntime-enclave-cc.yaml
@@ -77,23 +77,23 @@ spec:
             type: ""
           name: etc-systemd-system
         - hostPath:
-            path: /var/run/dbus
+            path: /var/run/dbus/system_bus_socket
             type: ""
           name: dbus
         - hostPath:
-            path: /run/systemd
+            path: /run/systemd/system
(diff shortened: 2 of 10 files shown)


wainersm (approved):
LGTM. IIUC it is going to fix the operator that got problem by changes on kata-deploy (http://jenkins.katacontainers.io/blue/organizations/jenkins/confidential-containers-ci/detail/confidential-containers-ci/16/pipeline)

fidencio:
> LGTM. IIUC it is going to fix the operator that got problem by changes on kata-deploy (http://jenkins.katacontainers.io/blue/organizations/jenkins/confidential-containers-ci/detail/confidential-containers-ci/16/pipeline)

Yep, thanks @wainersm!  I still need to figure out why the enclave-cc test is not passing though.

ryansavino:
> The SEV job failed because it hit pull limit to docker.io:
> 
> ```
> 13:28:22 TASK [Start a docker registry] *************************************************
> 13:28:24 fatal: [localhost]: FAILED! => {"changed": false, "msg": "Error pulling docker.io/library/registry - code: None message: toomanyrequests: You have reached your pull rate limit. You may increase the limit by authenticating and upgrading: [https://www.docker.com/increase-rate-limit"}](https://www.docker.com/increase-rate-limit%22%7D)
> ```
> 
> Have you seen that problem often, @ryansavino , @UnmeshDeodhar ?

I've retriggered the sev job. Looks like the docker hub rate limit affects the operator ci. Maybe we should consider pulling from GHCR or quay?

wainersm:
> > The SEV job failed because it hit pull limit to docker.io:
> > ```
> > 13:28:22 TASK [Start a docker registry] *************************************************
> > 13:28:24 fatal: [localhost]: FAILED! => {"changed": false, "msg": "Error pulling docker.io/library/registry - code: None message: toomanyrequests: You have reached your pull rate limit. You may increase the limit by authenticating and upgrading: [https://www.docker.com/increase-rate-limit"}](https://www.docker.com/increase-rate-limit%22%7D)
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Have you seen that problem often, @ryansavino , @UnmeshDeodhar ?
> 
> I've retriggered the sev job. Looks like the docker hub rate limit affects the operator ci. Maybe we should consider pulling from GHCR or quay?

I've opened an issue to track that issue: [tests/e2e: hitting docker.io/library/registry pull rate limit](https://github.com/confidential-containers/operator/issues/182) and @UnmeshDeodhar will be working on a fix.

mythi:
I debugged the enclave-cc failure and my conclusion is it will not work: `kind` nodes don't have dbus running so all `systemctl` commands are failing. Not sure why it worked with CentOS 7. It's using systemd-219 which is from 2015 so it could be `systemctl` was just very different then?

I managed to test a workaround but it has implications: when running without dbus, the pods deployed by the operator could run with `hostPID: true` to get access to PID 1. After that, `SYSTEMD_IGNORE_CHROOT=1 systemctl` works.

One observation while debugging this was that we restart systemd twice. I believe `pre/post-install` daemon could run unprivileged if it just copied containerd...

fidencio:
> I debugged the enclave-cc failure and my conclusion is it will not work: `kind` nodes don't have dbus running so all `systemctl` commands are failing. Not sure why it worked with CentOS 7. It's using systemd-219 which is from 2015 so it could be `systemctl` was just very different then?
> 
> I managed to test a workaround but it has implications: when running without dbus, the pods deployed by the operator could run with `hostPID: true` to get access to PID 1. After that, `SYSTEMD_IGNORE_CHROOT=1 systemctl` works.
> 
> One observation while debugging this was that we restart systemd twice. I believe `pre/post-install` daemon could run unprivileged if it just copied containerd...

Hmm.  This ia bad, sincerely bad.

mythi:
@fidencio I could try to ask kind to add dbus and explain our use-case (install custom runtimehandlers)

stevenhorsman:
The TDX failures are do to a pre-existing error: https://github.com/kata-containers/kata-containers/issues/6529

config/samples/ccruntime/default/kustomization.yaml
```diff
@@ -10,8 +10,8 @@ images:
 - name: quay.io/confidential-containers/container-engine-for-cc-payload
   newTag: 98a790e8abdcc06c4b629b290ebaa217bf82e305
 - name: quay.io/confidential-containers/runtime-payload
-  newName: quay.io/confidential-containers/runtime-payload
-  newTag: kata-containers-129e43d1ea5cca528b7b97234b7561219208a244-x86_64
+  newName: quay.io/confidential-containers/runtime-payload-ci
+  newTag: kata-containers-c9d6b46ea0923672a89e85a01e0d380249e3993f-x86_64
```
stevenhorsman, line 14: The runtime-payload tags are 13 and 14 days old now. Can we bump this to the latest and re-run the tests, or do you want to do that as a separate PR?

config/samples/enclave-cc/base/ccruntime-enclave-cc.yaml
```diff
@@ -77,23 +77,23 @@ spec:
...
             type: ""
           name: dbus
         - hostPath:
-            path: /run/systemd
+            path: /run/systemd/system
             type: ""
           name: systemd
     preInstall:
-      image: quay.io/confidential-containers/container-engine-for-cc-payload:1def978fdf6b0cb2383d43970c3dc484df3cad54
+      image: quay.io/confidential-containers/container-engine-for-cc-payload:98a790e8abdcc06c4b629b290ebaa217bf82e305
```
stevenhorsman, line 88: There is a new container engine image as well that is 2 days old rather than 17

stevenhorsman (commented):
A couple of questions about versions

config/samples/enclave-cc/base/ccruntime-enclave-cc.yaml
fidencio: The new container image was the one I used for testing the `chroot`, which didn't work.

config/samples/ccruntime/default/kustomization.yaml
fidencio: Hmmm.  I'd prefer having this one merged and unblock possible work, but I'm fine on waiting a little bit on that and actually have the TDX fix merged on Kata Containers first.

config/samples/enclave-cc/base/ccruntime-enclave-cc.yaml
stevenhorsman: Cool - we just want to be aware that latest is pointing at it (in the unlikely event that anyone is using it)

stevenhorsman (approved):
Thanks for bumping the runtime payload versions. Assuming the tests are passing, LGTM!

Issue #182, opened 2023-03-14
tests/e2e: hitting docker.io/library/registry pull rate limit


wainersm:
**Describe the bug**

The [SEV job](http://jenkins.katacontainers.io/job/confidential-containers-operator-main-ubuntu-20.04_sev-x86_64-containerd_kata-qemu-sev-PR) has failed due to hitting pull rate limit of the **docker.io/library/registry**:
```
13:27:03 TASK [Start a docker registry] *************************************************
13:27:10 fatal: [localhost]: FAILED! => {"changed": false, "msg": "Error pulling docker.io/library/registry - code: None message: toomanyrequests: You have reached your pull rate limit. You may increase the limit by authenticating and upgrading: [https://www.docker.com/increase-rate-limit"}](https://www.docker.com/increase-rate-limit%22%7D)
13:27:10 
```

That image is used on the e2e CI to start a locally registry: https://github.com/confidential-containers/operator/blob/main/tests/e2e/ansible/start_docker_registry.yml

Let's replace that image. Maybe push to quay.io?

wainersm:
@ryansavino @fitzthum @UnmeshDeodhar are you aware of that problem? Anyone working to solve it?

UnmeshDeodhar:
I can work on this @wainersm .

wainersm:
Cool, thanks @UnmeshDeodhar

ryansavino:
@wainersm we were discussing a bit in slack.

https://github.com/confidential-containers/operator/blob/a0fbbf40ad0848aee6c9ed90fbf7d001e50396c4/tests/e2e/ansible/start_docker_registry.yml#L28

Can we host a registry image from quay?

wainersm:
Hi @ryansavino ,

> @wainersm we were discussing a bit in slack.
> 
> https://github.com/confidential-containers/operator/blob/a0fbbf40ad0848aee6c9ed90fbf7d001e50396c4/tests/e2e/ansible/start_docker_registry.yml#L28
> 
> Can we host a registry image from quay?

Yesterday I spent some time looking for images that could replace [docker.io/library/registry:latest](http://docker.io/library/registry:latest) and I found none. So if there isn't any legal impediment (I don't think there is) then we could have a copy of that image on [quay.io](http://quay.io/), yes.

PR #184, opened 2023-03-22
Fix docker rate limit

UnmeshDeodhar:
Changing registry to a fixed version so that we will not hit the docker hub rate limit.

Fixes: #182


fidencio:
@UnmeshDeodhar, almost there, but it's still a very long line.

Mind to break it as shown below?
```
Switch to a fixed version of the registry containers, so a new version
is not pulled in every time a test runs, avoiding then hitting the
docker hub rate limit.
```

wainersm (approved):
Thanks @UnmeshDeodhar !

fidencio (approved):
lgtm, thanks @UnmeshDeodhar!

PR #187, opened 2023-03-29
Add kata-qemu-snp runtime class

jepio:
Following the example of kata-qemu-sev, add kata-qemu-snp where needed. This will result in a runtime class being defined and kata-deploy.sh configuring containerd and shim for kata-qemu-snp.

Fixes: kata-containers/kata-containers#6547

Depends on all the other PRs listed in #146

### DIFF
--- a/bundle/manifests/cc-operator.clusterserviceversion.yaml
+++ b/bundle/manifests/cc-operator.clusterserviceversion.yaml
@@ -214,6 +214,7 @@ metadata:
                 "kata-clh-tdx-eaa-kbc"
                 "kata-qemu",
                 "kata-qemu-sev",
+                "kata-qemu-snp",
                 "kata-qemu-tdx",
                 "kata-qemu-tdx-eaa-kbc",
                 "kata-remote"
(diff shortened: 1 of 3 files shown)


wainersm (approved):
I only grep'ed for `kata-qemu-sev` and found only those places to change.

wainersm:
@jepio would you like me to add a do-not-merge label to hold this until the remaining pieces aren't merged?

jepio:
yes please :)

fitzthum:
Time to revisit this? @jepio

fitzthum:
Since there is no SNP test, we probably don't want to add to this release. In the short term if people want to use SNP they'll have to dig it out as an experimental feature by making the changes in this PR manually to their CR yaml.

fidencio:
> Since there is no SNP test, we probably don't want to add to this release. In the short term if people want to use SNP they'll have to dig it out as an experimental feature by making the changes in this PR manually to their CR yaml.

Yep, this will be part of the next release when the SNP tests come.

It's good to remember, although orthogonal to this PR, that the SNP tests should and would be following the tests that other platforms are following, instead of build up atop of the current SEV tests, as we discussed in a meeting a few months ago.

ryansavino:
@jepio my bad, didn't see this. I duplicated these changes in #197.

fidencio:
Let me close this one as another PR with the same content was merged.

Issue #186, opened 2023-03-29
Add runtimeclass for peer pods


stevenhorsman:
Currently the peer-pods operator builds it's own runtime payload that overwrite the kata-configuration with it's remote version: [`ADD ${CONFIG}/configuration-remote.toml /opt/kata-artifacts/opt/confidential-containers/share/defaults/kata-containers/configuration.toml`](https://github.com/confidential-containers/cloud-api-adaptor/blob/0d6eda047e81eadbec3145407d2e55a6244f83cc/install/runtime-payload/Dockerfile#L13)

Now in https://github.com/kata-containers/kata-containers/issues/6349 we have added a remote configuration to kata, we want to add a separate runtime-class `kata-remote` for this, so we don't have to use the custom runtime payload anymore.

I think that what we need to do here is:
- Add `kata-remote` cc-operator.clusterserviceversion.yaml file
- Either:
  - Not modify the co-co ccruntime kustomization.yaml files that sets the runtimeClassNames, but instead add it to the runtimeClassNames in the peer pods version of the cc runtime
  - Modify the co-co ccruntime kustomization.yaml files that sets the runtimeClassNames to add kata-remote for both default and s390x. Update the coco ccruntime by adding the crio `installerVolumeMounts` and `installerVolumes` and then remove the peer-pods version of cc-runtime. I'm not sure what the implication of any peer pod kustomize might be here.
  
@jensfr @bpradipt - you know the operator much better than I do, so I'd love your input on what you think is the best approach here and whether we can remove the peer pod ccruntime altogether and re-use the coco one?

PR #189, opened 2023-03-30
remote: Add runtime class

stevenhorsman:
- Add new kata-remote runtime class for peer pods configuration support

Fixes: #186

### DIFF
--- a/bundle/manifests/cc-operator.clusterserviceversion.yaml
+++ b/bundle/manifests/cc-operator.clusterserviceversion.yaml
@@ -215,7 +215,8 @@ metadata:
                 "kata-qemu",
                 "kata-qemu-sev",
                 "kata-qemu-tdx",
-                "kata-qemu-tdx-eaa-kbc"
+                "kata-qemu-tdx-eaa-kbc",
+                "kata-remote"
               ],
               "uninstallCmd": [
                 "/opt/kata-artifacts/scripts/kata-deploy.sh",


wainersm:
Hi @stevenhorsman , I was wondering why you didn't change the config/samples/ccruntime/default/kustomization.yaml file but it is explained in the issue #186. You have opted for the first case, right?

I was also thinking that maybe we could have "kata-remote" and "kata-remote-caa" (the former defaulting to latest) because, in theory, we could have other remote implementations than CAA into the future. Or I am wrong and CAA is the only possible remote implementation?

stevenhorsman:
> Hi @stevenhorsman , I was wondering why you didn't change the config/samples/ccruntime/default/kustomization.yaml file but it is explained in the issue #186. You have opted for the first case, right?

Yeah - in the absence of any feedback I figured the safest option was to keep it separate as it's easier to merge later than unpick

> I was also thinking that maybe we could have "kata-remote" and "kata-remote-caa" (the former defaulting to latest) because, in theory, we could have other remote implementations than CAA into the future. Or I am wrong and CAA is the only possible remote implementation?

Yeah, I'm not against this and have thought about it before. Again I picked `kata-remote` out of simplicity when I created the kata configuration files (which I think have to match the runtimeclass), but I'm not against renaming this after 0.5 - just trying to limit the number of changes going in late in the release.

That said, if you feel strongly about doing these now then I can try and rework them.

wainersm:
> > Hi @stevenhorsman , I was wondering why you didn't change the config/samples/ccruntime/default/kustomization.yaml file but it is explained in the issue #186. You have opted for the first case, right?
> 
> Yeah - in the absence of any feedback I figured the safest option was to keep it separate as it's easier to merge later than unpick
> 

Got it!

> > I was also thinking that maybe we could have "kata-remote" and "kata-remote-caa" (the former defaulting to latest) because, in theory, we could have other remote implementations than CAA into the future. Or I am wrong and CAA is the only possible remote implementation?
> 
> Yeah, I'm not against this and have thought about it before. Again I picked `kata-remote` out of simplicity when I created the kata configuration files (which I think have to match the runtimeclass), but I'm not against renaming this after 0.5 - just trying to limit the number of changes going in late in the release.
>

It's fair.
 
> That said, if you feel strongly about doing these now then I can try and rework them.

Nope, let's keep it simple.

wainersm (approved):
Thanks @stevenhorsman

fidencio (approved):
lgtm, thanks @stevenhorsman!

stevenhorsman:
It looks like the CLH test did run and pass (http://jenkins.katacontainers.io/view/all/job/confidential-containers-operator-main-ubuntu-20.04-x86_64-containerd_kata-clh-PR/118/), but didn't report back for whatever reason, I think it's safe to merge this based on that.

PR #200, opened 2023-04-14
Update payloads to what will become v0.5.0

fidencio:
I'm updating the Enclave CC and Kata Containers payload.
There's no change on the pre-install payloads since the last bump.


stevenhorsman (approved):
LGTM.

sameo (approved):

PR #202, opened 2023-04-14
Add a sample ccruntime config for kata-remote (peer-pods)

bpradipt:


fidencio (approved):
lgtm, thanks @bpradipt!

config/samples/ccruntime/peer-pods/kustomization.yaml
```diff
@@ -0,0 +1,22 @@
...
+  newTag: 98a790e8abdcc06c4b629b290ebaa217bf82e305
+- name: quay.io/confidential-containers/runtime-payload
+  newName: quay.io/confidential-containers/runtime-payload-ci
+  newTag: kata-containers-e58ccb632fd95b7d83bfe85fba5ad79b1c03baaf-x86_64
+
+patches:
+- patch: |-
+    - op: replace
+      path: /spec/config/runtimeClassNames
+      value: ["kata-remote"]
```
stevenhorsman, line 20: I guess in the short-term we don't want to support mixed peer-pods and 'local pods' in the same environment, so only provide the remote runtime class?

bpradipt: Right.

bpradipt: Technically it's possible but as example for specifically peer-pods I felt it's better to have only the relevant runtimeclass

stevenhorsman: We might want to consider revisiting in future, but that's fine by me for now. Thanks for confirming.

stevenhorsman (approved):
LGTM. Thanks!

fidencio:
@stevenhorsman, I've retriggered the s390x job,

@ryansavino, @UnmeshDeodhar, @larrydewey, @fitzthum, we can just ignore the SNP test, as it's not officially part of this release, right?

Issue #213, opened 2023-05-25
Fix operator install from operatorhub


bpradipt:
Installation results in the following error

```
ubuntu@kubecon23eu:~$ kubectl -n operators describe ds cc-operator-pre-install-daemon
Name:           cc-operator-pre-install-daemon
Selector:       name=cc-operator-pre-install-daemon
Node-Selector:  node-role.kubernetes.io/worker=
Labels:         <none>
Annotations:    deprecated.daemonset.template.generation: 1
Desired Number of Nodes Scheduled: 0
Current Number of Nodes Scheduled: 0
Number of Nodes Scheduled with Up-to-date Pods: 0
Number of Nodes Scheduled with Available Pods: 0
Number of Nodes Misscheduled: 0
Pods Status:  0 Running / 0 Waiting / 0 Succeeded / 0 Failed
Pod Template:
  Labels:           name=cc-operator-pre-install-daemon
  Service Account:  cc-operator-controller-manager
  Containers:
   cc-runtime-pre-install-pod:
    Image:      quay.io/confidential-containers/container-engine-for-cc-payload:1def978fdf6b0cb2383d43970c3dc484df3cad54
    Port:       <none>
    Host Port:  <none>
    Command:
      /bin/sh
      -c
      /opt/confidential-containers-pre-install-artifacts/scripts/pre-install.sh
    Environment:
      NODE_NAME:   (v1:spec.nodeName)
    Mounts:
      /etc/systemd/system/ from etc-systemd-system (rw)
      /opt/confidential-containers/ from confidential-containers-artifacts (rw)
      /run/systemd from systemd (rw)
      /var/run/dbus from dbus (rw)
  Volumes:
   confidential-containers-artifacts:
    Type:          HostPath (bare host directory volume)
    Path:          /opt/confidential-containers/
    HostPathType:  DirectoryOrCreate
   etc-systemd-system:
    Type:          HostPath (bare host directory volume)
    Path:          /etc/systemd/system/
    HostPathType:  
   dbus:
    Type:          HostPath (bare host directory volume)
    Path:          /var/run/dbus
    HostPathType:  
   systemd:
    Type:          HostPath (bare host directory volume)
    Path:          /run/systemd
    HostPathType:  
Events:
  Type     Reason        Age                  From                  Message
  ----     ------        ----                 ----                  -------
[...]

PR #214, opened 2023-05-26
Add privileged label to the install namespace

bpradipt:
Fixes:#213

### DIFF
--- a/config/crd/bases/confidentialcontainers.org_ccruntimes.yaml
+++ b/config/crd/bases/confidentialcontainers.org_ccruntimes.yaml
@@ -771,10 +771,13 @@ spec:
                                         If the provisioner or an external controller
                                         can support the specified data source, it
                                         will create a new volume based on the contents
-                                        of the specified data source. If the AnyVolumeDataSource
-                                        feature gate is enabled, this field will always
-                                        have the same contents as the DataSourceRef
-                                        field.'
+                                        of the specified data source. When the AnyVolumeDataSource
+                                        feature gate is enabled, dataSource contents
+                                        will be copied to dataSourceRef, and dataSourceRef
+                                        contents will be copied to dataSource when
+                                        dataSourceRef.namespace is not specified.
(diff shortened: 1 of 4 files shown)


bpradipt:
@fidencio @stevenhorsman ptal

bpradipt:
Rebased the code to the latest

jensfr (approved):
lgtm, thank you @bpradipt

bpradipt:
@stevenhorsman The SNP test doesn't seem to work. Can we merge this ?

stevenhorsman:
> @stevenhorsman The SNP test doesn't seem to work. Can we merge this ?

SNP is still WIP, so shouldn't block anything. Whether we merge it now depends if it should be in, or out of the release.

bpradipt:
> > @stevenhorsman The SNP test doesn't seem to work. Can we merge this ?
> 
> SNP is still WIP, so shouldn't block anything. Whether we merge it now depends if it should be in, or out of the release.

For the next release, let's have the SNP as not mandatory so that we don't unnecessarily wait for CI to pass and delay merging. Anyway forced merging was also available which we should have used. Lost opportunity :-( 

We did discuss in the last call to have this PR merged so that people installing from operator hub can do it without any issues. But now may not be the right time and anyone installing from operator hub will be required to manually label the namespace.

Anyways let's hold this and get back to it after the release is cut.

stevenhorsman:
> For the next release, let's have the SNP as not mandatory so that we don't unnecessarily wait for CI to pass and delay merging.

Hey Pradipta, I might be missing something, but the SNP tests aren't required, so I don't think it blocks merging? Or maybe I have admin rights so I don't see that? If snp is showing as mandatory then it shouldn't be, so we need to ensure that doesn't happen again!

stevenhorsman:
Given that we've cut the release, I guess we can merge this now? especially if we want to do a special operator hub v0.6.1 we should do it sooner rather than later?

fitzthum:
Yeah let's merge as soon as the tests are good.

PR #215, opened 2023-06-02
Update components to test v0.6.0

stevenhorsman:
Update pre-install, kata-containers ans enclave-cc payloads to the latest ready for release testing

### DIFF
--- a/config/samples/ccruntime/default/kustomization.yaml
+++ b/config/samples/ccruntime/default/kustomization.yaml
@@ -8,7 +8,7 @@ resources:
 
 images:
 - name: quay.io/confidential-containers/container-engine-for-cc-payload
-  newTag: latest
+  newTag: 98a790e8abdcc06c4b629b290ebaa217bf82e305
 - name: quay.io/confidential-containers/runtime-payload
   newName: quay.io/confidential-containers/runtime-payload-ci
   newTag: kata-containers-latest
--- a/config/samples/ccruntime/default/kustomization.yaml
+++ b/config/samples/ccruntime/default/kustomization.yaml
@@ -11,7 +11,7 @@ images:
   newTag: 98a790e8abdcc06c4b629b290ebaa217bf82e305
 - name: quay.io/confidential-containers/runtime-payload
   newName: quay.io/confidential-containers/runtime-payload-ci
-  newTag: kata-containers-latest
+  newTag: kata-containers-947466ce98e5a2d7ecb3f91f919f83296f5ce019
 
 patches:
 - patch: |-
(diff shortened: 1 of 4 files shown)


stevenhorsman:
@mythi - hey Mikko, I noticed that the format of the enclave-CC payloads seems to be updated to include the kbc, so just wanted to double check with you that was correct and the versions I've picked are the ones you'd expect? Thanks!

fitzthum (approved):
LGTM

fitzthum:
tests not looking great

stevenhorsman:
> tests not looking great

Yeah - I might have done something wrong in this PR, but I'm not sure what that is. I think it might need manually checking.

fitzthum:
Maybe we should try going back to the previous `container-engine` build. The tests are all having a similar issue with `ERROR: cc-operator-daemon-install pod is not running`

stevenhorsman:
> Maybe we should try going back to the previous `container-engine` build. The tests are all having a similar issue with `ERROR: cc-operator-daemon-install pod is not running`

Yeah - I'll give that a try

stevenhorsman:
> Yeah - I'll give that a try. I've pushed it as a separate commit whilst we debug, but can fixup if that helps.

Okay - I'll replace fixup the container-engine as that seemed to be the issue!

fitzthum:
Yeah that's more like it. I'm a little shaky on how these `container-engine` containers actually get generated. I don't see a correspondence between the image tag and the commit hashes. The seemingly broken container-engine is a month newer than the most recent change in the repo.

stevenhorsman:
> Yeah that's more like it. I'm a little shaky on how these `container-engine` containers actually get generated. I don't see a correspondence between the image tag and the commit hashes. The seemingly broken container-engine is a month newer than the most recent change in the repo.

Yeah - I picked the newest tag from quay.io, as I hoped that matched what `latest` was, but `latest` looks like a bit of a mess, so we stick with this one and then try and look at https://github.com/confidential-containers/operator/pull/179 soon after release so see if that sheds any light? I think we just wait for Mikko to review this and it's looking good for all the picked versions for the release 😄

fitzthum:
Yeah that all sounds good to me. Looks like we're in great shape for v0.6.0 going into next week.

mythi (approved):

mythi:
> @mythi - hey Mikko, I noticed that the format of the enclave-CC payloads seems to be updated to include the kbc, so just wanted to double check with you that was correct and the versions I've picked are the ones you'd expect? Thanks!

Yes, adding `cc-kbc` and `sample-kbc` to enclave-cc tags is correct. Thanks for helping with this! Do we need to do anything for the `container-engine-for-cc-payload` image version?

stevenhorsman:
> Thanks for helping with this! Do we need to do anything for the `container-engine-for-cc-payload` image version?

No - as you might be able to tell from the comments above - I bumped the payload image version and it didn't work, so went back to last release version that enclave-cc already had, so we are all good there.

Issue #221, opened 2023-06-27
GPG error: public key is not found


Amulyam24:
This error occurred while creating a new containerd pre-install payload image
```
#0 19.35 deb [signed-by=/etc/apt/keyrings/kubernetes-archive-keyring.gpg] https://apt.kubernetes.io/ kubernetes-xenial main
| #0 19.87 Hit:2 http://archive.ubuntu.com/ubuntu focal InRelease
| #0 19.89 Hit:3 http://security.ubuntu.com/ubuntu focal-security InRelease
| #0 20.08 Hit:4 http://archive.ubuntu.com/ubuntu focal-updates InRelease
| #0 20.16 Get:1 https://packages.cloud.google.com/apt kubernetes-xenial InRelease [8993 B]
| #0 20.20 Err:1 https://packages.cloud.google.com/apt kubernetes-xenial InRelease
| #0 20.20   The following signatures couldn't be verified because the public key is not available: NO_PUBKEY B53DC80D13EDEF05
| #0 20.30 Hit:5 http://archive.ubuntu.com/ubuntu focal-backports InRelease
| #0 20.38 Reading package lists...
| #0 21.26 W: GPG error: https://packages.cloud.google.com/apt kubernetes-xenial InRelease: The following signatures couldn't be verified because the public key is not available: NO_PUBKEY B53DC80D13EDEF0
5
| #0 21.26 E: The repository 'https://apt.kubernetes.io/ kubernetes-xenial InRelease' is not signed.
```

jepio:
@Amulyam24 this can be closed as we now pull the binary directly, instead of relying on the mirror (and gpg).

Amulyam24:
Yes, this can be closed. Thanks!

PR #223, opened 2023-06-27
containerd: switch to the right gpg key url

Amulyam24:
The google url fails with an error of public key not available, hence switch it to the right one.

Fixes: #221


fidencio (approved):
lgtm, thanks @Amulyam24!

bpradipt (approved):
/lgtm
Thanks @Amulyam24

wainersm:
hi @Amulyam24 ! Even with changing the gpg key we were having problems to get the packaged kubectl installed in the pre-install image. This way I proposed a different approach (please see https://github.com/confidential-containers/operator/commit/9f2f377171abf50247e6d109b27fe61febe2125d) that got merged today. Therefore, this PR is not needed anymore. Can I close it?

PR #232, opened 2023-07-21
Update artifacts to release candidates built with tags

fitzthum:
Very similar to the last update, but these bundles were built using the tagged version of the subcomponents.

Step 17 of https://github.com/confidential-containers/confidential-containers/issues/111


fidencio (approved):
lgtm, assuming green tests!

wainersm (approved):

fidencio:
I am re-running TDX CLH.

fidencio:
> I am re-running TDX CLH.

And it's green now.

fidencio:
"s390x requeue has also failed on the 3rd test again. I'll try again, but I don't think we should let it hold us up as if we can pull and verify signed and encrypted images, then unsigned and unencrypted should be okay too!", from Steven on https://cloud-native.slack.com/archives/C039JSH0807/p1689962125719039?thread_ts=1689257407.179079&cid=C039JSH0807

Merging then.

PR #233, opened 2023-07-21
Update artifacts to v0.7.0 release bundles

fitzthum:
These are the bundles corresponding to the v0.7.0 tags for CCv0 and enclave-cc. 

Step 23 or https://github.com/confidential-containers/confidential-containers/issues/111

### DIFF
--- a/config/samples/ccruntime/default/kustomization.yaml
+++ b/config/samples/ccruntime/default/kustomization.yaml
@@ -10,8 +10,8 @@ images:
 - name: quay.io/confidential-containers/container-engine-for-cc-payload
   newTag: c80c3e7392c1562841e75072d227e07f98cab2bd
 - name: quay.io/confidential-containers/runtime-payload
-  newName: quay.io/confidential-containers/runtime-payload-ci
-  newTag: kata-containers-f5420e5cf198eb54585dad9fb63eb2eaaff3e5a8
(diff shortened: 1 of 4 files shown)


wainersm (approved):
The runtime-payload was generated by the workflow: https://github.com/kata-containers/kata-containers/actions/runs/5626077123

wainersm:
The s390x job fail:
```
18:44:01 [init] Using Kubernetes version: v1.24.0
18:44:01 [preflight] Running pre-flight checks
18:44:01 error execution phase preflight: [preflight] Some fatal errors occurred:
18:44:01 	[ERROR Port-6443]: Port 6443 is in use
18:44:01 [preflight] If you know what you are doing, you can make a check non-fatal with `--ignore-preflight-errors=...`
```

It suggests that a previous job did not clean up the node correctly. @BbolroC could you please have a look?

BbolroC:
I've placed a clean-up code in the jenkins job. A port 6443 should be closed before the test starts. But it seems the 3rd test `Test can pull an unencrypted image inside the guest` is failing. I have no idea for why. It is over midnight in my place. I will have a look at the test next Monday. 😉

BbolroC:
Good news, the 3rd test has passed when I run the test on another node. I think this has something to do with an infra. Let's see if the whole test ends up with going green.

stevenhorsman (approved):
LGTM

PR #236, opened 2023-07-25
Fix node selector label for ccRuntime to a well known label

shivamerla:


fidencio:
@shivamerla, please, take a look at https://github.com/confidential-containers/operator/commit/abd7952648d3c90646c1f81f9bff4693130a5cc9

That was the reason we moved away from "node-role".

Please, feel free to close the PR in case the reference is enough for that,

fidencio:
I'm closing this one, please, if the explanation above was not enough, feel free to re-open it.

shivamerla:
@fidencio yes, i was not aware of such limitation. makes sense now, thanks. Unfortunately tools like `kubeadm` still apply that node label, so we would need to change the default matchLabel with `ccRuntime`.

PR #243, opened 2023-08-10
crd: Expose "UsingNFD" to the CcInstallConfig

fidencio:
Let's have a explicit toggle that allows users to specify whether they're relying on NFD or not.

In case they're not relying on NFD, business as usual.  In case they are, then the runtime payload will be able to create a specific NodeFeatureRule, adapt the podOverhead of the runtime class, and make sure the keys book-keeping is correctly done.

Right now, on the Kata Containers side of the things, only TDX is capable of doing so, but it's easy to expand in case other TEEs want to do the same.

### DIFF
--- a/api/v1beta1/ccruntime_types.go
+++ b/api/v1beta1/ccruntime_types.go
@@ -172,6 +172,10 @@ type CcInstallConfig struct {
 	// +optional
 	Debug bool `json:"debug,omitempty"`
 
+	// This specified whether NodeFeatureDiscovery is being used or not, and relies on its feataures in case it is.
+	// +optional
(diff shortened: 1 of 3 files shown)


fidencio:
A few notes for the reviewers.

1. This work depends on a PR that's not yet merged on the Kata Containers side: https://github.com/kata-containers/kata-containers/pull/7469
2. I've decided to use an explicit toggle as rules, on the Kata Containers side, were only introduced for TDX
3. We may hit the case where someone is using NFD, but still may want to have their own rules being used (and then they can manually do it, as they're most likely doing Today)

fidencio:
JFYI, this one will be material for the v0.9.0 release, not for this one.
Thanks for the understanding.

PR #237, opened 2023-07-25
Update commands to expand RELEASE_VERSION

shivamerla:

PR #246, opened 2023-08-11
docs: standardizing variable calls

PamplonaCarlos:
At the documentation, the variables are called in different ways, but executing in a "$<>" standard, it appears to occur the error:
````
sh: Syntax error: newline expected
````

Changing the way it's made by "${}" solves the problem.


fidencio:
This is a dup of https://github.com/confidential-containers/operator/pull/237, and we'll prioritise that one as it came if first.
Thanks for the contribution @PamplonaCarlos!

Issue #163, opened 2023-02-27
Update operator-sdk to latest version


bpradipt:
Creating this tracker issue to evaluate what is required to update the operator-sdk version.

bpradipt:
@caavery are you also bumping the CoCo operator sdk version as part of https://github.com/confidential-containers/cloud-api-adaptor/issues/872 ?

ariel-adam:
@caavery is this on track for release 0.7.0 (feature freeze 12th of July) or should we move it to 0.8.0?

PR #227, opened 2023-07-20
go.mod: update dependencies + drop custom kind node image

mythi:

PR #249, opened 2023-08-16
Update go.mod to include peerpods. Add peerpod controllers to operator. 

caavery:
Fixes: #163
Fixes: # https://github.com/confidential-containers/cloud-api-adaptor/issues/872

This is a draft PR as there are some outstanding issues with adding the peerpod controllers:
-  Changes in Dockerfile and config/manager/manager.yaml were made to accommodate libvirt and the ssh key for peerpod-ctrl as libvirt was the test platform.
-  The peerpod web-hook modifications are not done and therefore not included.
-  Make uninstall produces some operator errors although everything cleans up. The operator log is attached.

[operator-uninstall.log](https://github.com/confidential-containers/operator/files/12360070/operator-uninstall.log)


caavery:
@bpradipt @snir911 @stevenhorsman

mythi:
See also #227

caavery:
@mythi 
I am anticipating having to rebase if thats what you are indicating.
Cathy

mythi:
> @mythi I am anticipating having to rebase if thats what you are indicating. Cathy

I misunderstood the intent of this PR. It's not moving APIs to k8s.io v1.28 as suggested by 40a04ca title but about something else so no issues there.

caavery:
commit 40a04ca updates go.mod to operator sdk 1.28.

mythi:
> commit [40a04ca](https://github.com/confidential-containers/operator/commit/40a04caa16b5e40696717df69373bfc9f188f3c2) updates go.mod to operator sdk 1.28. If the title is misleading I can change it.

commit title talks about k8s APIs. I don't quite follow the part where operator-sdk version has a dependency to what k8s APIs we use. Therefore, I think 40a04ca is not needed at all.

caavery:
Due to adding the peerpod modules ( at go 1.19 ) it was necessary to upgrade the operator go.mod to 1.19 as well as I was receiving build errors.  This also necessitated syncing to a comparable operator sdk version using go 1.19  the latest of which was 1.28.

This is the same practice as go.mod commit:
 
commit f4980bf4430f135154bc294d41027f568c4a4b7e
Author: Pradipta Banerjee <pradipta.banerjee@gmail.com>
Date:   Fri Aug 26 05:09:44 2022 +0000

    Update code and deps to align with sdk v1.23 and k8s api 0.24.2
    
    K8s api and golang version used should be in sync with the operator-sdk version.
    For operator-sdk 1.23 that is used for this PR, the corresponding golang and k8s api
    versions can be found here - https://github.com/operator-framework/operator-sdk/blob/v1.23.0/go.mod#L37

mythi:
> This also necessitated syncing to a comparable operator sdk version using go 1.19 the latest of which was 1.28.

Go 1.21.0 is already available so 1.19 is end-of-life.

caavery:
That could be done as a separate issue to follow. Right now the operator is a go.18

Dockerfile
```diff
@@ -19,14 +19,21 @@ COPY main.go main.go
 COPY api/ api/
 COPY controllers/ controllers/
 
+# The container running the operator/controllers needs the libvirt libraries as
+# as the peerpod-ctrl needs to dynamically link with libvirt
+RUN apt-get update -y && apt-get install -y libvirt-dev
+
 # Build
-RUN CGO_ENABLED=0 GOOS=linux go build -a -o manager main.go
+RUN CGO_ENABLED=1 GOOS=linux go build -tags=libvirt -a -o manager main.go
```
snir911, line 27: all of these changes should be related to peerpod-ctrl commit no?

config/manager/manager.yaml
```diff
@@ -23,12 +23,22 @@ spec:
...
     spec:
       securityContext:
-        runAsNonRoot: true
+        runAsNonRoot: false
+      volumes:
+      - name: ssh
+        secret:
+          defaultMode: 384
+          optional: true
+          secretName: ssh-key-secret
```
snir911, line 32: peerpod-ctrl still doesn't work with ssh authentication right? i think you should have comment on this and point to https://github.com/confidential-containers/cloud-api-adaptor/issues/1241

snir911 (commented):
Thanks! can you explain how to test it? 
I added some comments, mostly splitting the peerpod-ctrl required changes to its own are needed i think, overall lgtm
cc @jensfr for reviewing the peerpodconfig part

Dockerfile
caavery: Yes they are. I will move them to the other peerpod-ctrl commit.

caavery:
@snir911 

I can add how to test it but do you mean here in the comments section?

api/v1beta1/ccruntime_types.go
```diff
@@ -38,6 +38,11 @@ type CcRuntimeSpec struct {
 	RuntimeName CcRuntimeName `json:"runtimeName"`
 
 	Config CcInstallConfig `json:"config"`
+
+	// EnablePeerPods is used to transparently create pods on a remote system.
+	// +optional
+	// +kubebuilder:default:=false
+	EnablePeerPods bool `json:"enablePeerPods"`
```
mythi, line 45: Why is this API change necessary? AFAICS users can deploy peerpods CRs separately and the peerpods controller does its job.

config/manager/manager.yaml
```diff
@@ -23,12 +23,22 @@ spec:
         control-plane: controller-manager
     spec:
       securityContext:
-        runAsNonRoot: true
+        runAsNonRoot: false
+      volumes:
```
mythi, line 27: Add these via overlay patches so that non-peerpod users don't get them

main.go
```diff
@@ -104,6 +114,26 @@ func main() {
 		setupLog.Error(err, "unable to create controller", "controller", "CcRuntime")
 		os.Exit(1)
 	}
+
+	if err = (&peerpodconfigcontrollers.PeerPodConfigReconciler{
```
mythi, line 118: Only enable these reconcilers through a commandline arg so that all users don't get them

caavery:
@mythi 
Thanks for the feedback!! I'll be working on it and let you know if I have any questions.

config/manager/manager.yaml
caavery: I'm not entirely sure I understand what you have in mind. Are you talking about changing make deploy based on some peerpod env variable to keep this out of manager.yaml?

mythi: yes, that was my suggestion

api/v1beta1/ccruntime_types.go
caavery: I can remove it from CcRuntimeSpec  but it is used in a couple of places in ccruntime_controller.go. 
I am assuming I need some sort of test there. Perhaps it can be derived from enabling peerpods in the command line argument.

https://github.com/confidential-containers/operator/blob/fed10fdfd1405bd171bdab6992e145ba5c8ab7dd/controllers/ccruntime_controller.go#L443
https://github.com/confidential-containers/operator/blob/fed10fdfd1405bd171bdab6992e145ba5c8ab7dd/controllers/ccruntime_controller.go#L253

caavery: PS sorry it took so long to get back to you. I was on PTO.

mythi: > but it is used in a couple of places in ccruntime_controller.go.

It's not obvious to me why ccruntime controller needs to be aware of peerpods resources.

I'd think you can do something like:

```
resources:
- ../base
- peers-pods-config.yaml
```

in  `config/samples/ccruntime/peer-pods/kustomization.yaml`. Peer Pods controller will then reconcile resources it owns.

caavery: I am pretty new to this stuff and I was following the sandboxed operator implementation.  If peerpods is not enabled falling into this code without a check would produce errors wouldn't it?

if r.ccRuntime.Spec.EnablePeerPods {
						r.Log.Info("Disable peerpods")
						// We are explicitly ignoring any errors in peerpodconfig removal as
						// these can be removed manually if needed and this is not in the critical path
						// of operator functionality
						_ = r.disablePeerPods()
					}
Thanks

mythi: > falling into this code without a check would produce errors wouldn't it?

As far as I can understand, code modifications in `ccruntime_controller.go` are not needed because adding PeerPods config is same as doing:

```
apiVersion: confidentialcontainers.org/v1alpha1
kind: PeerPodConfig
metadata:
  name: peerpodconfig-coco
spec:
  cloudSecretName: "peer-pods-secrets"
  configMapName:   "peer-pods-cm",
  limit: "10"
```

as part of the kustomization like I suggested above.

caavery: OK I can certainly try it out. Will let you know. Thanks

caavery: I added to kustomization.yaml
 resources:
- ../base
- peerpodsconfig.yaml

And my peerpodsconfig.yaml
apiVersion: confidentialcontainers.org/v1alpha1
kind: PeerPodConfig
metadata:
  name: peerpodconfig-coco
spec:
  cloudSecretName: "peer-pods-secrets"
  configMapName:   "peer-pods-cm",
  limit: "10"

However I am getting the error
kubectl create -k config/samples/ccruntime/peer-pods/
error: accumulating resources: accumulating resources from 'peerpodsconfig.yaml': MalformedYAMLError: yaml: line 5: did not find expected key in File: peerpodsconfig.yaml

caavery: After fixing the wayward comma in peerspodsconfig.yaml it created the peerpodconfig controller
cc-operator-controller-manager-68b5979488-g75kd   2/2     Running   0          3h11m
cc-operator-daemon-install-kmk6g                  1/1     Running   0          57s
cc-operator-pre-install-daemon-nsxjd              1/1     Running   0          59s
peerpodconfig-ctrl-caa-daemon-bppq5               1/1     Running   0          58s

So far so good!

mythi: @caavery oops, yes sorry about my syntax errors

caavery: It was like hiding in plain sight. Totally happy for the info! Now I have a bit of naming suffix/prefix '-peer-pod' build up:
kcli@virtlab723 operator]$  k get peerpodconfigs
NAME                           AGE
peerpodconfig-coco-peer-pods   164m

config/manager/manager.yaml
caavery: I have devised a way to do this but kustomize is exhibiting unexpected behavior at least in my opinion.
My patches go in and it all works but the config/default/kustomization.yaml file gets corrupted so that anything below the patches list gets deleted regardless of if I am doing a patch add or a patch delete. So all the WEBHOOK and CERTMANGER comments go away.

```
# Mount the controller config file for loading manager configurations
# through a ComponentConfig type
#- manager_config_patch.yaml

# To enable the peerpod controllers
- path: peerpods/peerpods-namespace.yaml
- path: peerpods/enable-peerpods.yaml
  target:
    group: apps
    kind: Deployment
    name: controller-manager
    version: v1

# [WEBHOOK] To enable webhook, uncomment all the sections with [WEBHOOK] prefix including the one in
# crd/kustomization.yaml
#- manager_webhook_patch.yaml

# [CERTMANAGER] To enable cert-manager, uncomment all sections with 'CERTMANAGER'.
# Uncomment 'CERTMANAGER' sections in crd/kustomization.yaml to enable the CA injection in the admission webhooks.
# 'CERTMANAGER' needs to be enabled to use ca injection
#- webhookcainjection_patch.yaml

```
```
Makefile 

deploy: manifests kustomize ## Deploy controller to the K8s cluster specified in ~/.kube/config.
        cd config/manager && $(KUSTOMIZE) edit set image controller=${IMG}
ifneq (, $(PEERPODS))
        @echo Peerpods will be deployed and added to the manager config yamls.
        cd config/default && $(KUSTOMIZE) edit add patch --kind Deployment --version  v1 --name controller-manager --group apps  --path peerpods/enable-peerpods.yaml
        cd config/default && $(KUSTOMIZE) edit add patch --path peerpods/peerpods-namespace.yaml
endif
        $(KUSTOMIZE) build config/default | kubectl apply -f -
```
The other alternative is to not make this dynamic patching and instead leave the patch paths in as part of the yaml and do what seems to be pretty standard which is comment them all out with the instructions

 [PEERPODS] To enable peerpods, uncomment all the sections with [PEERPODS]  etc, etc, etc

caavery: I'm going to try the latest version of kustomize to see if that helps

caavery:
@mythi @snir911 @bpradipt @jensfr 

I believe I have addressed all the feedback. I also created a separate Dockerfile for peerpods so non-peerpod users would not be affected. I however could not overcome the kustomize edit issue described in 
https://github.com/confidential-containers/operator/pull/249#discussion_r1329227738
so I instead used the standard manual uncomment technique employed by WEBHOOK and CERTMANAGER. 

The jenkins-ci-tdx-x86_64 check is failing for some reason but when I click on the details link I get a 'this site can't be reached error'.

Thanks,
Cathy

mythi:
> I believe I have addressed all the feedback. I also created a separate Dockerfile for peerpods so non-peerpod users would not be

The suggestions I made earlier are addressed so I'm OK. One more question still: I find it problematic to have the CRDs _copied_ to this repo. Similar to how we pull peerpod Go dependencies, would it be possible to use kustomize for the CRDs?

Something like:
```
apiVersion: kustomize.config.k8s.io/v1beta1
kind: Kustomization
resources:
- "https://github.com/confidential-containers/cloud-api-adaptor/tree/main/peerpod-ctrl/config/crd?ref=v0.7.0"
```

Would `operator-sdk` bundle/csv creation work?

caavery:
@mythi 

OK so you just don't want them copied or do you want them commented out   
config/crd/kustomization.yaml with instructions on enablement?  Yeah I don't know about bundle/cvs. New territory.  I can take a look. Do you understand what is going on with the  jenkins-ci-tdx-x86_64 test.? Says it failed but won't tell me why.

Thanks

mythi:
> OK so you just don't want them copied and pulled instead and do you want them commented out in

I was just saying copying them here creates maintenance overhead (because they cannot be generated from the source in this repo)

caavery:
K got it.

caavery:
I have added the upstream pull of the peerpod CRDs and removed the local copy.

config/samples/ccruntime/peer-pods/peerpodsconfig.yaml
```diff
@@ -0,0 +1,9 @@
+apiVersion: confidentialcontainers.org/v1alpha1
+kind: PeerPodConfig
+metadata:
+  name: coco-config
+  namespace: confidential-containers-system
+spec:
+  cloudSecretName: "peer-pods-secret"
+  configMapName:   "peer-pods-cm"
+  limit: "10"
```
jensfr, line 9: I think that default limit was changed to 1 in cloud-api-adaptor. Should we use the same here?

controllers/ccruntime_controller.go
```diff
@@ -57,13 +57,27 @@ type CcRuntimeReconciler struct {
 	Namespace string
 }
 
+const (
+	peerpodConfigCRName = "peerpodconfig-coco"
+	DEFAULT_PEER_PODS   = "10"
```
jensfr, line 62: Is this required? and if it is, should it be 1 as well?

jensfr (approved):
I left a few comments about the default limit value for peer pods, I think it should be changed to match the default value in the peer-pod-config controller code. But it's not  urgent and could be solved in a follow-up PR as well. So I'm going to approve this PR. 

Thank you @caavery!

config/default/kustomization.yaml
```diff
@@ -36,6 +37,18 @@ patches:
 # through a ComponentConfig type
 #- manager_config_patch.yaml
 
+# [PEERPODS] To enable the peerpod controllers uncomment this section.
```
mythi, line 40: would it make sense to have `config/default-with-peerpods` overlay that adds all the peerpods configs on top of `default`. It'd prefer to have the rbac settings maintained part of that overlay too.

controllers/ccruntime_controller.go
```diff
@@ -57,13 +57,27 @@ type CcRuntimeReconciler struct {
...
 //+kubebuilder:rbac:groups="",resources=namespaces,verbs=get;update
+//+kubebuilder:rbac:groups="",resources=nodes/status,verbs=patch
+//+kubebuilder:rbac:groups=confidentialcontainers.org,resources=peerpodconfigs,verbs=get;list;watch;create;update;patch;delete
+//+kubebuilder:rbac:groups=confidentialcontainers.org,resources=peerpodconfigs/status,verbs=get;update;patch
+//+kubebuilder:rbac:groups=confidentialcontainers.org,resources=peerpodconfigs/finalizers,verbs=update
+//+kubebuilder:rbac:groups=confidentialcontainers.org,resources=peerpods,verbs=get;list;watch;create;update;patch;delete
+//+kubebuilder:rbac:groups=confidentialcontainers.org,resources=peerpods/status,verbs=get;update;patch
+//+kubebuilder:rbac:groups=confidentialcontainers.org,resources=peerpods/finalizers,verbs=update
+//+kubebuilder:rbac:groups="",resources=configmaps,verbs=create;get;update;list;watch
+//+kubebuilder:rbac:groups="",resources=secrets,verbs=create;get;update;list;watch
```
mythi, line 80: these tags do not belong here.

caavery: @mythi 
OK you mean that I should get rid of the peerpods subdirectory and combine all the peerpod config yamls into one patch.yaml. So config/default/kustomization.yaml would contain:

# [PEERPODS] To enable the peerpod controllers uncomment this section.
#- manager_peerpods_patch.yaml

As far as the rbac permissions goes you want those moved from the go file to this manager_peerpods_patch.yaml as well. Could you show me an example as I am not familiar with implementing rbac in this way.

mythi: I was thinking something like `config/overlays/peerpods/kustomization.yaml` with

```
resources:
- ../../default
- <rbac from cloud-api-adator>
- <peerpods crds from cloud-api-adaptor>
- any_other_yaml_needed_here.yaml
```

caavery: Whats to prevent it from being pulled in when peerpods is not selected?

caavery: And you want me to move the crds out of config/crd/kustomization.yaml which I just changed to an upstream pull?

mythi: > Whats to prevent this new yaml from being pulled in when peerpods is not selected?

`kubectl -k config/overlays/peerpods` would not be used by those who don't want/need peerpods. They continue to use `config/default` without even noticing peerpods was added.

> And you want me to move the crds out of config/crd/kustomization.yaml which I recently changed to an upstream pull as well?

Yes, that's not a good place for those. They can be part of the peerpods overlay. My thinking is like this: peerpods APIs have their own lifecycle in that other repository. The APIs (CRDs), the corresponding controllers, necessary RBAC rules etc. are maintained elsewhere and this repo just consumes releases from peerpods repo. The same way `go.mod` dependencies are updated for the controller code, the corresponding yamls are pulled via tagged kustomize rules.

caavery: OK I'll give it a whirl.

caavery: @mythi @jensfr 
I've been able to separate out most things to an overlays/peerpod directory with 3 subdirectories ( crd, rbac, default ). One reason I created separate directories is the crds are applied at install and not deploy with the other yamls. I am having difficulty with the rbac separation. Specifically pulling from the upstream confidential-containers/cloud-api-adaptor/peerpodconfig-ctrl and confidential-containers/cloud-api-adaptor/peerpod-ctrl rbac directories. These files have openshift stuff in them and are not generic. What I would like to do is go ahead and place all the rbac files I need for the peerpod controllers in the rbac overlays subdirectory and create a CAA Issues request for a generic isolation of the peerpod controller's rbacs.

mythi: @caavery did you push the updates for review?

fidencio:
JFYI, this one will be material for the v0.9.0 release, not for this one.
Thanks for the understanding.

controllers/ccruntime_controller.go
caavery: @mythi @jensfr 
I have pushed the patches addressing the latest review comments with the exception of importing the peerpod rbacs as proposed in https://github.com/confidential-containers/operator/pull/249#discussion_r1365754689

mythi: The latest version looks good to me, thanks @caavery!

caavery: @mythi 

Thank you very much for all your help! BTW the PR doesn't seem to be registering your approval. 

Take Care,
Cathy

mythi (approved):
LGTM.

I did not look too closely but I can see the structure is well in place so that existing operator users are not impacted and peer-pods users can "opt-in" for what they need.

bpradipt:
@jensfr @mythi unfortunately this PR was never merged and now it needs a rebase. Since Cathy is no longer working on this, I'm helping to get this rebased. I'll be opening up another PR using the same code base as I can't modify this PR

PR #250, opened 2023-08-22
crd: Add and use RuntimeClasses struct to associate a runtime to a snapshotter

fidencio:
This is the Operator counter part of the following Kata Containers PR https://github.com/kata-containers/kata-containers/pull/7715

The idea here is to expose a way that users can set a snapshotter to be used with the containerd runtime handler.

The reason I'm marking this feature, right now, as "ExperimentalFeature" is a way to indicate this will be changed, or even dropped, in the coming releases, as Kata Containers will actually expose a way to map a specific runtime handler to a specific snapshotter, rather than "set this snapshotter for all the runtime handlers" as currently done.

This will require folks using a new enough version of containerd (v1.7+), but I do believe this will be the case anyways with the work going on as part of removing the dependency on the containerd fork.


BbolroC (approved):
LGTM, thanks @fidencio !

fidencio:
I am thinking out loud here, but for experimental features, we may just want to pass those as environment variables till things get actually stable.  Making this PR invalid as it is.

@bpradipt, @jensfr, WDYT?

bpradipt:
We never really discussed how to handle experimental features. For this specific case, since there could be different snapshotter configuration for a specific RuntimeClass, I was thinking should we change the RuntimeClassNames to a struct and add specifics like name, snapshotter config and any other required configs ?
 
https://github.com/confidential-containers/operator/blob/cce4aac93430de54f306647265c2b068f0db1ab7/api/v1beta1/ccruntime_types.go#L163

bpradipt:
> I still have to learn how to properly patch a struct, but that's for Tomorrow.
> 
> @bpradipt, thanks a ton for the suggestion, and I hope breaking backwards compat in v1beta1 is not a huge deal.

Yeah, till we release v1, we can be flexible

api/v1beta1/ccruntime_types.go
```diff
@@ -313,6 +313,14 @@ type FailedNodeStatus struct {
 	Error string `json:"error"`
 }
 
+// RuntimeClass holds the name and the snapshotter to be used by a runtime class
+type RuntimeClass struct {
+	// Name of the runtime class
+	Name string `json:"name"`
+	// The snapshotter to be used by the runtime class
+	Snapshotter string `json:"snapshotter"`
```
bpradipt, line 321: Any other inputs needed for the snapshotter ?

fidencio: To be more clear, this is the counter part change on the Kata Containers side: https://github.com/kata-containers/kata-containers/pull/7715/files

config/samples/ccruntime/ssh-demo/kustomization.yaml
```diff
@@ -15,8 +15,14 @@ images:
 patches:
 - patch: |-
     - op: replace
-      path: /spec/config/runtimeClassNames
-      value: ["kata", "kata-clh", "kata-qemu"]
+      path: /spec/config/runtimeClasses
+      value:
+      - name: "kata"
+        snapshotter: "overlayfs"
+      - name: "kata-qemu"
```
bpradipt, line 22: typo? kata-clh ?

fidencio: definitely. :-)

fidencio: Fixed, thanks @bpradipt!

bpradipt (approved):
/lgtm
Thanks @fidencio

stevenhorsman (approved):
LGTM. Thanks!

PR #253, opened 2023-08-25
pre-install-payload: Make its creation more generic

fidencio:
This is a ground work that will make things way simpler for https://github.com/confidential-containers/operator/pull/251


fidencio:
This PR is rebased atop of https://github.com/confidential-containers/operator/pull/179

install/pre-install-payload/Dockerfile
```diff
@@ -1,25 +1,47 @@
-FROM ubuntu:22.04
+#### Confidential Containers forked containerd
+
+FROM alpine:3.18 as coco-containerd-binary-downloader
```
wainersm, line 3: Just a nit: instead of repeat `FROM alpine:3.18` for the three images, you can create a `FROM alpine:3.18 as base`  and the subsequent images takes `FROM base`. On that base image you can even update the image like `apt get update` (don't know if Alpine has the same update mechanism)

fidencio: Perfect, will do!

wainersm:
@fidencio I reviewed the last 5 commits, no more nits :)

Actually, maybe one more... as you are changing names... `pre-install-post-uninstall` is somewhat confusing. I understand the idea, it contain the code to pre-install and post-uninstall but what if we call it `node-requirements` or `coco-requirements` or `coco-deps` or `setup-pre-coco`... but I don't want to hold your progress on this work if it will take much efforts to change.

fidencio:
> it contain the code to pre-install and post-uninstall but what if we call it `node-requirements` or `coco-requirements` or `coco-deps` or `setup-pre-coco`

I just renamed it to `reqs-payload`.

wainersm:
ah, @fidencio , I just remembered... there is a automation to build the (old) pre-install image on PR merger (https://github.com/confidential-containers/operator/blob/main/.github/workflows/pre-install-image-publish-on-merge.yml) that will need changes. I can send an follow up PR with those things as I was the one who introduced the workflow on first place.

wainersm (approved):
Thanks @fidencio ! But it is a pity we can't kill the pre-install image :(

wainersm:
> > ah, @fidencio , I just remembered... there is a automation to build the (old) pre-install image on PR merger (https://github.com/confidential-containers/operator/blob/main/.github/workflows/pre-install-image-publish-on-merge.yml) that will need changes. I can send an follow up PR with those things as I was the one who introduced the workflow on first place.
> 
> This is covered here already: https://github.com/confidential-containers/operator/pull/253/files#diff-0688aecd94d71cf248e44a407562596e3f8337eebf8d324e419403927a100089R44

oh, true, I missed that.

Issue #254, opened 2023-08-29
quay.io/confidential-containers/reqs-payload cannot run reqs-deploy.sh: No such file or directory


mattarnoatibm:
**Describe the bug**
The `quay.io/confidential-containers/reqs-payload` image in the Kustomization files is pinned at an image that doesn't contain the reqs-deploy.sh script, so when that image runs as part of the peer pod operator install the `cc-operator-pre-install-daemon` pod goes into a crash loop

**To Reproduce**
1. Deploy the peer pods operator on a Kubernetes cluster (I used an IKS cluster)
 
**Describe the results you expected**
The cc-operator-pre-install-daemon pods are created and run, and don't go into a crash loop

**Describe the results you received:**
The cc-operator-pre-install-daemon pods go into a crash loop with the error
`/opt/confidential-containers-pre-install-artifacts/scripts/pre-install.sh: line 3: /opt/confidential-containers-pre-install-artifacts/scripts/reqs-deploy.sh: No such file or directory`

**Additional context**
The `The quay.io/confidential-containers/reqs-payload:4ad0978b8e9f1acb857fe2c76f1494c55bf28e05` image doesn't contain a script named `reqs-deploy.sh` in `/opt/confidential-containers-pre-install-artifacts/scripts/` so `/opt/confidential-containers-pre-install-artifacts/scripts/pre-install.sh`, which tries to run `/opt/confidential-containers-pre-install-artifacts/scripts/reqs-deploy.sh`, errors.

In the `4ad0978b8e9f1acb857fe2c76f1494c55bf28e05` image, the content that should be in `reqs-deploy.sh` seems to be in a script `pre-install-post-uninstall-deploy.sh` instead.

The `quay.io/confidential-containers/reqs-payload` image tagged with `e45d4e84c3ce4ae116f3f4d6c123c4829606026f` and currently tagged with `latest` contains the `reqs-deploy.sh` in the expected place.

PR #255, opened 2023-08-29
pre-install-payload: Use latest tagged runtime-payload image

mattarnoatibm:
The tags for the quay.io/confidential-containers/reqs-payload image were not set to the latest tag and were set to the tag of an image that went into a crash loop when you ran it as part of the operator installation.

Fixes: #254

### DIFF
--- a/config/samples/ccruntime/default/kustomization.yaml
+++ b/config/samples/ccruntime/default/kustomization.yaml
@@ -8,7 +8,7 @@ resources:
 
 images:
 - name: quay.io/confidential-containers/reqs-payload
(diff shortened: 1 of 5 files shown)


fidencio (approved):
lgtm, thanks @mattarnoatibm!

Just one comment that I think using latest is fine, but we need to tag to a specific version during the release.

mattarnoatibm:
I'll update this PR to set the reqs-payload to e45d4e84c3ce4ae116f3f4d6c123c4829606026f. Thanks for the feedback Steve and Fabiano.

PR #263, opened 2023-10-09
do-not-merge: This will be used to test the kata-containers CI related stuff

fidencio:

PR #267, opened 2023-10-20
Use the nydus-snapshotter by default

fidencio:
This PR is a rework of https://github.com/confidential-containers/operator/pull/263, but I don't want to lose that public reference of what's been done.

Please, take a look at each  commit for more info on what it's doing and why.


install/pre-install-payload/Makefile
```diff
@@ -1,11 +1,13 @@
 COCO_CONTAINERD_VERSION = 1.6.8.2
 OFFICIAL_CONTAINERD_VERSION = 1.7.0
 VFIO_GPU_CONTAINERD_VERSION = 1.7.0.0
+NYDUS_SNAPSHOTTER_VERSION = v0.13.2
```
wainersm, line 4: Hi @fidencio,

v0.13.2 contains the fixes? Because on CCv0 we are still on v0.12.0 (https://github.com/kata-containers/kata-containers/blob/CCv0/versions.yaml#L305)

install/pre-install-payload/scripts/reqs-deploy.sh
```diff
@@ -53,6 +54,24 @@ function install_vfio_gpu_containerd_artefacts() {
 	install_containerd_artefacts "vfio-gpu"
 }
 
+function install_nydus_snapshotter_artefacts() {
+	echo "Copying nydus-snapshotter artifacts onto host"
+
+	install -D -m 755 ${artifacts_dir}/opt/confidential-containers/bin/containerd-nydus-grpc /opt/confidential-containers/bin/containerd-nydus-grpc
```
wainersm, line 60: In the end, the /usr/local/bin/containerd-nydus-grpc -> /opt/confidential-containers/bin/containerd-nydus-grpc link not required, right?

install/pre-install-payload/scripts/reqs-deploy.sh
```diff
@@ -87,6 +110,24 @@ function uninstall_containerd_artefacts() {
 	fi
 }
 
+function uninstall_nydus_snapshotter_artefacts() {
+	remove_nydus_snapshotter_from_containerd
+
+	host_systemctl disable nydus-snapshotter.service
```
wainersm, line 116: Shouldn't stop the snapshotter service before? or passing `---now` to systemctl's disable should do the trick too.

wainersm: @fidencio finished my review. This is the only question preventing me of approving it.

install/pre-install-payload/Makefile
fidencio: We're just using the latest here, and that's a known to be working version.
We shouldn't be lacking behind because of CCv0.

install/pre-install-payload/scripts/reqs-deploy.sh
fidencio: No, `containerd-nydus-grpc` doesn't have to be in `/usr/local/bin` as we call it directly to start the systemd service.

install/pre-install-payload/scripts/reqs-deploy.sh
fidencio: > Shouldn't stop the snapshotter service before? or passing ---now to systemctl's disable should do the trick too.

Very nice catch!

install/pre-install-payload/Makefile
wainersm: Agree. I'd like to double-check though.

wainersm (approved):
@fidencio amazing work as always! Ready to land this IMO.

install/pre-install-payload/scripts/reqs-deploy.sh
```diff
@@ -87,6 +110,27 @@ function uninstall_containerd_artefacts() {
 	fi
 }
 
+function uninstall_nydus_snapshotter_artefacts() {
```
fitzthum, line 113: Where is this called from?

fitzthum (commented):
Tested a bit on my node. I think we might not be uninstalling nydus at all (see comment). Otherwise looks good.

install/pre-install-payload/scripts/reqs-deploy.sh
fidencio: This should've been called from `uninstall_artifacts`, but I missed it.
Very nice catch!

install/pre-install-payload/scripts/reqs-deploy.sh
```diff
@@ -99,6 +143,47 @@ function restart_systemd_service() {
 	host_systemctl restart "${container_engine}"
 }
 
+function configure_nydus_snapshotter_for_containerd() {
+	echo "configure nydus snapshotter for containerd"
+
+	if [ ! -f "$containerd_config" ]; then
+		die "failed to find containerd config"
```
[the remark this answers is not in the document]
fidencio: We do that on the kata-deploy payload, but to be completely honest ... I'm not a big fond of doing that, and let me try to explain why.

1. We'd need to mount /usr/bin from host, which we currently do not do, to ensure `containerd` is available.
2. We'd need to ensure the containerd version used to generate this is the version being used on the host, which is easy in the case where we install containerd, but not so easy to ensure running on CSPs nodes

I must even say that we may need to rething on using that on the kata-containers side as well.


OTOH, I'm defintely **not entirely against** of doing that, that can be done, but I just think that we need something slightly more robust than simply calling `containerd config default`.

Does it make sense?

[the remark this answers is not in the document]
fitzthum: Yeah seems like more trouble than it is worth at least for this PR.

fitzthum (approved):
LGTM

We might see more SEV/SNP failures if other PRs (or the baseline) have run given that they will leave the nodes in a bad state. Should work if we rerun a few times tho.

fidencio:
> We might see more SEV/SNP failures if other PRs (or the baseline) have run given that they will leave the nodes in a bad state. Should work if we rerun a few times tho.

Cool, thanks!

Right now I'm re-working the uninstall as it made the jobs related to uninstall / reinstall break.

controllers/ccruntime_controller.go
```diff
@@ -226,6 +226,20 @@ func (r *CcRuntimeReconciler) processCcRuntimeDeleteRequest() (ctrl.Result, erro
 			if r.ccRuntime.Spec.Config.PostUninstall.Image == "" {
 				controllerutil.RemoveFinalizer(r.ccRuntime, RuntimeConfigFinalizer)
 			} else if r.ccRuntime.Spec.Config.PostUninstall.Image != "" {
+				// FXIME: This should be treated in a better way, as just having the sleep
```
bpradipt, line 229: nit: FXIME-> FIXME
This looks reasonable. We'll focus on a proper fix in the next release

bpradipt (approved):
/lgtm
Thanks @fidencio

controllers/ccruntime_controller.go
fidencio: I will fix the fixme in the next iteration, just in order to save some trees and unblock peer-pods folks. :-)

PR #268, opened 2023-10-24
tests/e2e: describe all pods on teardown

wainersm:
On operator_tests.bats's teardown let's print the description of all pods in the confidential-containers namespace to help on debug fails.


wainersm:
This one is part of https://github.com/confidential-containers/operator/pull/267 so closing it...

Issue #271, opened 2023-10-27
Take into consideration that /etc/containerd/config.toml may not exists


fidencio:
Containerd can run fine if the default config doesn't exist, and we should take this into consideration.

This has been discussed here: https://github.com/confidential-containers/operator/pull/267#discussion_r1370633947

PR #273, opened 2023-10-27
pre-install: Create containerd config it it doesn't exist

fidencio:
We're relying here on containerd being on /usr/local/bin/, which we do have access to, in order to create such config.

We're creating the config as the first thing on the script, as we don't want to get into the situation where a configuration was added / removed and we end up with that in the system due to us using a newer version of containerd than the one present in the system.

We may face more issues in the future, such as:
* containerd is only in /usr/bin/
  * We can start exposing this as a host path mount, but let's not do it unless it becomes necessary
* when CRI-O is supported, this file should only be created in case containerd is used

However, for now, let's go with this approach and avoid users hitting a possible error due to the lack of contained configuration.

Fixes: #271

### DIFF
--- a/install/pre-install-payload/scripts/reqs-deploy.sh
+++ b/install/pre-install-payload/scripts/reqs-deploy.sh
@@ -149,10 +149,6 @@ function restart_systemd_service() {
 function configure_nydus_snapshotter_for_containerd() {
 	echo "configure nydus snapshotter for containerd"
 
-	if [ ! -f "$containerd_config" ]; then
-		die "failed to find containerd config"
-	fi
-
 	containerd_imports_path="/etc/containerd/config.toml.d"
 
 	echo "Create ${containerd_imports_path}"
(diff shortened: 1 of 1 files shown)


fidencio:
All the BM tests on jenkins are failing due to:
```
13:59:51 stratovirt.go:971:20: ineffectual assignment to err (ineffassign)
13:59:51 	s.virtiofsDaemon, err = s.createVirtiofsDaemon(hypervisorConfig.SharedPath)
13:59:51 	      
```

stevenhorsman (commented):
I think you are already aware of this, but when I building the image and testing out the code it created the `/etc/containerd/config.toml`, but it had no content

fidencio:
> I think you are already aware of this, but when I building the image and testing out the code it created the `/etc/containerd/config.toml`, but it had no content

That was happening due to the lack of `gcompat` installed in the pre-reqs base image, which is based on `alpine`.

fitzthum (approved):
LGTM

PR #274, opened 2023-10-27
ccruntime: Set `INSTALL_OFFICIAL_CONTAINERD` to false

fidencio:
The majority of the managed kubernetes solutions are already relying on
a new enough (v1.7+) version of containerd.

/cc @wainersm, we'd need to adapt CI for this, and I'd need your help to do so.

PR #277, opened 2023-10-30
config: set operator payloads for v0.8.0 release

portersrc:
This covers steps 4 and 5 from the release checklist [here](https://github.com/confidential-containers/confidential-containers/issues/145)

### DIFF
--- a/config/samples/ccruntime/default/kustomization.yaml
+++ b/config/samples/ccruntime/default/kustomization.yaml
@@ -11,7 +11,7 @@ images:
   newTag: fad6ef4607e929c2d30d16dc8c62fbb29881a44c
 - name: quay.io/confidential-containers/runtime-payload
   newName: quay.io/confidential-containers/runtime-payload-ci
-  newTag: kata-containers-latest
+  newTag: kata-containers-424de1cbfa4e1da9ecf9a56b1d1e1a11a4f339cd
 
 patches:
 - patch: |-
--- a/config/samples/ccruntime/default/kustomization.yaml
+++ b/config/samples/ccruntime/default/kustomization.yaml
@@ -8,7 +8,7 @@ resources:
 
 images:
 - name: quay.io/confidential-containers/reqs-payload
-  newTag: latest
+  newTag: fad6ef4607e929c2d30d16dc8c62fbb29881a44c
 - name: quay.io/confidential-containers/runtime-payload
   newName: quay.io/confidential-containers/runtime-payload-ci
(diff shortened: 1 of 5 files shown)


ChengyuZhu6:
tdx test is passed: http://10.112.240.228:8080/job/confidential-containers-operator-main-centos8stream-x86_64-containerd_kata-qemu-tdx-PR/588/

fidencio:
@portersrc, just one tip, mind the commit message.

Here's a little bit of git etiquette that you can get some inspiration from https://gist.github.com/robertpainsi/b632364184e70900af4ab688decf6f53

config/samples/ccruntime/default/kustomization.yaml
```diff
@@ -8,10 +8,10 @@ resources:
 
 images:
 - name: quay.io/confidential-containers/reqs-payload
-  newTag: latest
+  newTag: fad6ef4607e929c2d30d16dc8c62fbb29881a44c
```
stevenhorsman, line 11: Don't we need https://github.com/confidential-containers/operator/pull/274 to be merged first before we take the reqs-payload release?

fitzthum: hm good point. i guess the question is whether we can continue the release in parallel (i.e. merge this, tag the subprojects, and pick up the new reqs/operator changes when we do the next bundle update (step 13)). i am fine with waiting for #274, but that will serialize things a bit. wdyt @fidencio

stevenhorsman: If the plan is to cycle back around and re-bump this once the operator PRs are finished then I'm okay with that, I just wanted to ensure that people were aware that we haven't finished yet, so some of this work will probably need to be redone.

fitzthum: Yeah #280 also will change things. I think it is reasonable to pick up these changes later. It seems likely that a bundle that works with the current operator would also work with those PRs merged. Not ideal to have both things moving at once.

stevenhorsman (approved):
Based on the understanding that some of the tags will need to be updated before the release is cut, I'm okay with this change.

config/samples/ccruntime/default/kustomization.yaml
fidencio: I just merged this one, there are PRs coming on the Operator side; there may be PRs coming on the peer-pods side as well, but I'm fine on having kata-containers side of the things **not moing**.

Issue #278, opened 2023-10-31
Artefacts are being left behind after uninstall


fidencio:
```
ubuntu@operator:~$ tree /opt/confidential-containers/
/opt/confidential-containers/
└── share
    └── nydus-snapshotter
        └── config-coco-guest-pulling.toml

2 directories, 1 file
```

^^^ this is not being cleaned up after the uninstall

PR #280, opened 2023-10-31
pre-install: pre-install: Remove nydus-snapshotter config

fidencio:
We've been leaving the nydus-snapshotter config behind because we were
trying to remove it from a path that was changed from the first
iteration of the nydus-snapshotter addition, but ended up not being
noticed during review.

While here, let's also make sure to entirely remove
`/opt/confidential-containers/share`, as nydus-snapshotter is the only
bit of code using that.

Fixes: #278

While here let's also pre-remove ctr cleanup of nydus snapshotters

This has proven to be useless as it is, as `ctr` doesn't have access to
the `/var/run/{containerd,containerd-nydus}` to actually be able to
perform any operation.

We could, obviously, expose those to the operator, but I'm very much
against adding even more host mounts to the operator, unless this is
strictly needed, and right now I don't think it is (and I'm fine to be
proven wrong in the future ;-)).

### DIFF
--- a/install/pre-install-payload/scripts/reqs-deploy.sh
+++ b/install/pre-install-payload/scripts/reqs-deploy.sh
@@ -126,7 +126,12 @@ function uninstall_nydus_snapshotter_artefacts() {
 	rm -f /opt/confidential-containers/bin/containerd-nydus-grpc
 	rm -f /opt/confidential-containers/bin/nydus-overlayfs
 	rm -f /usr/local/bin/nydus-overlayfs
-	rm -f /opt/confidential-containers/share/remote-snapshotter/config-coco-guest-pulling.toml
+	rm -f /opt/confidential-containers/share/nydus-snapshotter/config-coco-guest-pulling.toml
+
+	# We can do this here as we're sure that only the nydus-snapshotter is
+	# installing something in the /opt/confidential-containers/share
+	# directory
+	rm -rf /opt/confidential-containers/share
 	rm -rf /var/lib/containerd-nydus/*
 }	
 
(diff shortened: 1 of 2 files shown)


fitzthum (approved):
LGTM. We might want to run the tests twice here because this is a change to the uninstall stuff. 

Also, @ryansavino looks like we need some cleanup on the SEV/SNP node.

wainersm (approved):
@fidencio lgtm

install/pre-install-payload/scripts/reqs-deploy.sh
```diff
@@ -112,8 +116,8 @@ function uninstall_containerd_artefacts() {
 
 function uninstall_nydus_snapshotter_artefacts() {
 	if host_systemctl list-units | grep -q nydus-snapshotter; then
-		for i in `ctr -n k8s.io snapshot --snapshotter nydus list | grep -v KEY | cut -d' ' -f1`; do
-			ctr -n k8s.io snapshot --snapshotter nydus rm $i || true
+		for i in `host_ctr -n k8s.io snapshot --snapshotter nydus list | grep -v KEY | cut -d' ' -f1`; do
+			host_ctr -n k8s.io snapshot --snapshotter nydus rm $i || true
```
wainersm, line 120: @fidencio I was about to ask you if there is a situation where `ctr` is not installed, but this `|| true` will handle any error.

wainersm (approved):
very clever way of solving it @fidencio !

fitzthum (approved):
LGTM

fidencio:
I'm going ahead and merging this one, as the changes will most likely help with the AMD / Intel CIs.

PR #205, opened 2023-04-14
release: add update_ccruntime_images.sh

wainersm:
Created the update_ccruntime_images.sh script which should help on update the runtime-payload and pre-install images accross the ccruntimes variants.

Fixes #204
Signed-off-by: Wainer dos Santos Moschetta <wainersm@redhat.com>

PR #282, opened 2023-11-06
config: update operator payloads for v0.8.0 release

portersrc:
This covers step 13 from the release checklist [here](https://github.com/confidential-containers/confidential-containers/issues/145)

### DIFF
--- a/config/samples/enclave-cc/base/ccruntime-enclave-cc.yaml
+++ b/config/samples/enclave-cc/base/ccruntime-enclave-cc.yaml
@@ -9,7 +9,7 @@ spec:
       node.kubernetes.io/worker: ""
   config:
     installType: bundle
-    payloadImage: quay.io/confidential-containers/runtime-payload-ci:enclave-cc-HW-cc-kbc-b58c59dc72a0b10ae3dd4436a206fd54a820c9f7
+    payloadImage: quay.io/confidential-containers/runtime-payload-ci:enclave-cc-HW-cc-kbc-ed979ef2952d8b20c2f0abee3d32a3617d2c57bb
     installDoneLabel:
       confidentialcontainers.org/enclave-cc: "true"
     uninstallDoneLabel:
--- a/config/samples/enclave-cc/sim/kustomization.yaml
+++ b/config/samples/enclave-cc/sim/kustomization.yaml
@@ -7,4 +7,4 @@ images:
 - name: quay.io/confidential-containers/runtime-payload-ci
   newTag: enclave-cc-SIM-sample-kbc-ed979ef2952d8b20c2f0abee3d32a3617d2c57bb
 - name: quay.io/confidential-containers/reqs-payload
-  newTag: fad6ef4607e929c2d30d16dc8c62fbb29881a44c
+  newTag: c31205eb31d8a4623ffc0c9fa316e3c91e337023
--- a/config/samples/enclave-cc/sim/kustomization.yaml
+++ b/config/samples/enclave-cc/sim/kustomization.yaml
@@ -5,6 +5,6 @@ nameSuffix: -sgx-mode-sim
 
 images:
 - name: quay.io/confidential-containers/runtime-payload-ci
-  newTag: enclave-cc-SIM-sample-kbc-b58c59dc72a0b10ae3dd4436a206fd54a820c9f7
+  newTag: enclave-cc-SIM-sample-kbc-ed979ef2952d8b20c2f0abee3d32a3617d2c57bb
 - name: quay.io/confidential-containers/reqs-payload
   newTag: fad6ef4607e929c2d30d16dc8c62fbb29881a44c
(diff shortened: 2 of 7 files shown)


config/samples/enclave-cc/base/ccruntime-enclave-cc.yaml
```diff
@@ -9,7 +9,7 @@ spec:
       node.kubernetes.io/worker: ""
   config:
     installType: bundle
-    payloadImage: quay.io/confidential-containers/runtime-payload-ci:enclave-cc-HW-cc-kbc-b58c59dc72a0b10ae3dd4436a206fd54a820c9f7
+    payloadImage: quay.io/confidential-containers/runtime-payload-ci:enclave-cc-HW-cc-kbc-753599f76ef201968438b15f01d7b37d49cc7c5c
```
fidencio, line 12: For this one ...

config/samples/enclave-cc/sim/kustomization.yaml
```diff
@@ -5,6 +5,6 @@ nameSuffix: -sgx-mode-sim
 
 images:
 - name: quay.io/confidential-containers/runtime-payload-ci
-  newTag: enclave-cc-SIM-sample-kbc-b58c59dc72a0b10ae3dd4436a206fd54a820c9f7
+  newTag: enclave-cc-SIM-sample-kbc-753599f76ef201968438b15f01d7b37d49cc7c5c
```
fidencio, line 8: ... and this one, could you use ed979ef2952d8b20c2f0abee3d32a3617d2c57bb tag?

enclave-cc merged a bump on Occlum Today, and we want to test with that image.

fidencio (changes requested):
@portersrc, please, just take a look at the comments related to enclave-cc.

portersrc:
> @portersrc, please, just take a look at the comments related to enclave-cc.

I bumped it; it looks like that enclave-cc bundle is still building as of this writing though

fidencio (approved):
lgtm, assuming the tests will pass.

Thanks, @portersrc!

tests/e2e/operator.sh
```diff
@@ -66,7 +66,7 @@ build_pre_install_img() {
 handle_older_containerd() {
         command -v containerd >/dev/null || return
         local version
-        version=$(containerd -v | awk '{ print $3 }')
+        version=$(containerd -v | awk '{ print $3 }' | sed 's/^v//')
```
wainersm, line 69: Thanks for the fix, @portersrc !

fitzthum (commented):
LGTM

fitzthum (approved):

stevenhorsman (approved):
LGTM. Thanks!

fidencio:
> @portersrc any special reason to not pin the s390x payload? -> https://github.com/confidential-containers/operator/blob/main/config/samples/ccruntime/s390x/kustomization.yaml

Oh, that's something that was overlook, I guess, nice catch!

wainersm:
> > @portersrc any special reason to not pin the s390x payload? -> https://github.com/confidential-containers/operator/blob/main/config/samples/ccruntime/s390x/kustomization.yaml
> 
> Oh, that's something that was overlook, I guess, nice catch!

Also we are using the CI payload instead of `quay.io/confidential-containers/runtime-payload`

wainersm:
@portersrc you also need to bump the hash of pre-reqs for peer-pods too

Good opportunity to advertise this: https://github.com/confidential-containers/operator/pull/205 (perhaps someone "cof-cof Chris cof-cof" would like to take over that script).

fitzthum:
> Also we are using the CI payload instead of quay.io/confidential-containers/runtime-payload

I think the idea is to switch to the runtime-payload bundle in step 19 after we add the release tags for kata and enclave-cc. In the past we have sometimes combined this step with that one, though, and we could do that again if people think it would save time.

wainersm:
> > Also we are using the CI payload instead of quay.io/confidential-containers/runtime-payload
> 
> I think the idea is to switch to the runtime-payload bundle in step 19 after we add the release tags for kata and enclave-cc. In the past we have sometimes combined this step with that one, though, and we could do that again if people think it would save time.

Good point @fitzthum , I forgot kata isn't tagged yet. IMHO, anything to speed up **this** release is welcomed, so I am okay to switch the payload on this PR to run the CI jobs one more time

portersrc:
Thanks to @wainersm for catching these two things: pinning s390x runtime-payload-ci and also pinning peer-pods' reqs-payload.
These are incorporated into the latest push.

fitzthum:
@wainersm my hunch for now is to do another PR for step 19 just to make sure we don't confuse ourselves. I don't think that will take too much longer, but I leave it to @portersrc to decide.

wainersm (approved):
@portersrc now it looks okay to me. Thanks!

fitzthum:
Let's give these test one more chance to behave

PR #284, opened 2023-11-09
reqs: handle the absence of disable_snapshot_annotations

wainersm:
The `disable_snapshot_annotations` should be set to `false` in the containerd's configuration.toml in order to make the remote snapshotter work. The reqs-deploy.sh currently handles only the case where the property is already present on the configuration file, i.e., it will switch from `true` to `false` only. However, on some nodes like in AWS EKS and Azure AKS, `disable_snapshot_annotations` simply doesn't exist so the current `sed` takes no effect (and the current containerd's behavior is to disable annotations).

This changed the handler so that on the absence of `disable_snapshot_annotations`, it is appended to the `[plugins."io.containerd.grpc.v1.cri".containerd]` section.

Fix for **issue 2** in https://github.com/kata-containers/kata-containers/issues/8407


wainersm:
I'm marking this as do-not-merge because I still need to handle the uninstall case, i.e., we need to somehow mark that the line was added then remove it.

I was thinking in literally put a commit on the line just above the `disable_snapshot_annotations` line. For example:

```
# disable_snapshot_annotations - GENERATED BY COCO OPERATOR
disable_snapshot_annotations = false
```

Then on uninstall it will know the line should be removed. Yes, ugly hack... any other ideas?

wainersm:
@fitzthum @fidencio @stevenhorsman it would be nice to have your review and hear ideas for the uninstall problem.

stevenhorsman:
> Yes, ugly hack... any other ideas?

I agree that's it's not great, but I can't think of anything better and I've definitely written worse hacks!

wainersm:
With the generated comment, it will append the property like this:

```
[plugins."io.containerd.grpc.v1.cri".containerd]
# disable_snapshot_annotations - DO NOT EDIT. GENERATED BY CoCo OPERATOR
disable_snapshot_annotations = false
  default_runtime_name = "runc"
  discard_unpacked_layers = true
```

On uninstall, delete both lines resulting in
```
[plugins."io.containerd.grpc.v1.cri".containerd]
  default_runtime_name = "runc"
  discard_unpacked_layers = true
```

Here is the diff:
```
diff --git a/install/pre-install-payload/scripts/reqs-deploy.sh b/install/pre-install-payload/scripts/reqs
-deploy.sh
index a76345a..5185f13 100755
--- a/install/pre-install-payload/scripts/reqs-deploy.sh
+++ b/install/pre-install-payload/scripts/reqs-deploy.sh
@@ -12,6 +12,8 @@ INSTALL_NYDUS_SNAPSHOTTER=${INSTALL_NYDUS_SNAPSHOTTER:-true}
 containerd_config="/etc/containerd/config.toml"
 artifacts_dir="/opt/confidential-containers-pre-install-artifacts"
 
+readonly snapshot_annotations_marker="disable_snapshot_annotations - DO NOT EDIT. GENERATED BY CoCo OPERA
TOR"
+
 die() {
        msg="$*"
        echo "ERROR: $msg" >&2
@@ -192,7 +194,7 @@ EOF
        else
                # In case the property does not exist, let's append it to the
                # [plugins."io.containerd.grpc.v1.cri".containerd] section.
-               sed -i '/\[plugins\..*\.containerd\]/adisable_snapshot_annotations = false' \
+               sed -i '/\[plugins\..*\.containerd\]/a'"# ${snapshot_annotations_marker}"'\ndisable_snapsh
ot_annotations = false' \
                        "${containerd_config}"
        fi
 
@@ -211,7 +213,14 @@ function remove_nydus_snapshotter_from_containerd() {
        sed -i -e "s|\"${containerd_imports_path}/nydus-snapshotter.toml\"||g" ${containerd_config}
        sed -i -e "s|, ]|]|g" ${containerd_config}
 
-       sed -i -e "s|disable_snapshot_annotations = false|disable_snapshot_annotations = true|" ${containe
rd_config}
+       if grep -q "${snapshot_annotations_marker}" "${containerd_config}"; then
+               sed -i '/'"${snapshot_annotations_marker}"'/d' \
+                       "${containerd_config}"
+               sed -i '/disable_snapshot_annotations = false/d' \
+                       "${containerd_config}"
+       else
+               sed -i -e "s|disable_snapshot_annotations = false|disable_snapshot_annotations = true|" ${
containerd_config}
+       fi
 }
 
 label_node() {
```

wainersm:
@mkulke you can test this PR (including the logic to uninstall nydus) with `quay.io/wainersm/reqs-payload:latest`

That image was built as:
```
$ cd install/pre-install-payload
$ make registry=quay.io/wainersm/reqs-payload reqs-image
```

install/pre-install-payload/scripts/reqs-deploy.sh
```diff
@@ -183,7 +183,23 @@ EOF
...
+	# Annotations should be passed down to the remote snapshotter in order to
+	# make it work. This can be done by setting `disable_snapshot_annotations = false`
+	# in the containerd's config.toml.
+	if grep -q 'disable_snapshot_annotations' "$containerd_config"; then
+		sed -i -e "s|disable_snapshot_annotations = true|disable_snapshot_annotations = false|" \
+			"${containerd_config}"
+	else
+		# In case the property does not exist, let's append it to the
+		# [plugins."io.containerd.grpc.v1.cri".containerd] section.
+		sed -i '/\[plugins\..*\.containerd\]/adisable_snapshot_annotations = false' \
```
portersrc, line 195: is this a typo? `adisable_snapshot_annotations` vs. `disable_snapshot_annotations`?

wainersm: @portersrc the next char after '/' can be a command to sed. In this case, '/a' tells to Append the following string (i.e disable_snapshot_annotations) ;)

Thanks for the careful review!

mkulke:
> @mkulke you can test this PR (including the logic to uninstall nydus) with `quay.io/wainersm/reqs-payload:latest`
> 
> That image was built as:
> 
> ```
> $ cd install/pre-install-payload
> $ make registry=quay.io/wainersm/reqs-payload reqs-image
> ```

Thanks. I built reqs-payload images from your branch and used referenced them in the kustomize config for ccruntime/peer-pods. I can confirm it works, the line is being added to `/etc/containerd/config.toml`

mkulke:
regarding the uninstall part, I think that's the best we can do. Reverting a configuration change is a brittle endeavour, we don't own the config, so a user/tool could do any sort of changes between a coco install and uninstall, which could mess with our undo logic.

install/pre-install-payload/scripts/reqs-deploy.sh
mkulke: fwiw: if we want to keep the indent this should work w/ nested capture groups:

```bash
sed -e 's|\(\(\s*\)\[plugins.\"io.containerd.grpc.v1.cri\".containerd\]\)|\1\n\2  ohai=true|' /etc/containerd/config.toml | grep -C1 ohai
    [plugins."io.containerd.grpc.v1.cri".containerd]
      ohai=true
      default_runtime_name = "runc"
 ```

wainersm:
> > @mkulke you can test this PR (including the logic to uninstall nydus) with `quay.io/wainersm/reqs-payload:latest`
> > That image was built as:
> > ```
> > $ cd install/pre-install-payload
> > $ make registry=quay.io/wainersm/reqs-payload reqs-image
> > ```
> 
> Thanks. I built reqs-payload images from your branch and used referenced them in the kustomize config for ccruntime/peer-pods. I can confirm it works, the line is being added to `/etc/containerd/config.toml`

Great! Let me incorporate the uninstall logic on this PR then run the CI.

wainersm:
> regarding the uninstall part, I think that's the best we can do. Reverting a configuration change is a brittle endeavour, we don't own the config, so a user/tool could do any sort of changes between a coco install and uninstall, which could mess with our undo logic.

Exactly. Let's see if for long term we find a solution that doesn't need to touch containerd's files.

wainersm:
Let the fun begin:

/test

fitzthum:
@wainersm, maybe you want to add a commit to [bump the operator version](https://github.com/confidential-containers/operator/blob/main/config/release/kustomization.yaml#L7) to this PR to save us a step?

wainersm:
@fitzthum added the commit to bump the release version

wainersm:
let the fun begin (2):

/test

fitzthum:
I guess we will need one more PR after this to update the reqs payload.

fitzthum (approved):
LGTM. Maybe the version bump should actually have gone in the next PR, but not a big deal.

fidencio (approved):
lgtm, thanks @wainersm and @mkulke for the back and forth and exhaustive tests on this!

Issue #294, opened 2023-11-27
E2e test for s390x should be migrated to GHA


BbolroC:
**Is your feature request related to a problem? Please describe.**
As part of the migration of the CI jobs on Jenkins to GHA, CI jobs for s390x named

- `confidential-containers-operator-main-ubuntu-20.04-s390x-containerd_kata-qemu-PR`
- `confidential-containers-operator-main-ubuntu-20.04-s390x-containerd_kata-qemu-baseline`

should be migrated.

PR #295, opened 2023-11-28
tests/e2e: Migrate e2e test for s390x to GHA

BbolroC:
This PR is to migrate the following e2e tests for s390x in Jenkins to GHA:

- confidential-containers-operator-main-ubuntu-20.04-s390x-containerd_kata-qemu-baseline
- confidential-containers-operator-main-ubuntu-20.04-s390x-containerd_kata-qemu-PR
- confidential-containers-operator-main-ubuntu-20.04-s390x-SE-daily

It will be staying in a `work-in-progress` state and the workflow structure will be changed based on that of `x86_64`. (i.e. a decision could be made when https://github.com/confidential-containers/operator/pull/260 is finalised.)

The workflow itself has been tested several times

- https://github.com/BbolroC/cc-operator/actions/runs/7017907526
- https://github.com/BbolroC/cc-operator/actions/runs/7017396678
- https://github.com/BbolroC/cc-operator/actions/runs/7011926398

The rationale for the `{pre,post}` action in the workflow is as follows:

1. A self-hosted runner for s390x cannot be instantiated and used instantly on request, rather should be ready/running 24/7. It is inevitable to institute a hook for the management over test runs.
2. The management script could be committed to the repo, but it wouldn't be flexible enough to tackle any environmental issues on the runner.

Of course, an introduction of the actions makes it hard to incorporate workflows for different architectures into one via `matrix` because `if:` for the selective run for an architecture (e.g. `if: ${{ matrix.arch == 's390x' }}`) conflicts with `if: always()` which is used for the post action. I look forward to any feedback for those two actions.

There are 2 things which must be prepared before the merge:

1. Registration for a self-hosted runner for s390x
2. Configuration of a secret named `REGISTRY_CREDENTIAL_ENCODED`.

Thanks.

Fixes: #294

### DIFF
--- a/.github/workflows/ccruntime_e2e.yaml
+++ b/.github/workflows/ccruntime_e2e.yaml
@@ -1,8 +1,11 @@
 name: ccruntime e2e tests
 on:
-  pull_request:
-    branches:
-      - main
+  workflow_call:
+    inputs:
+      target-branch:
+        required: false
+        type: string
+        default: ""
 
 permissions:
   contents: read
@@ -13,11 +16,31 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        runtimeclass: ["kata-qemu", "kata-clh"]
-        instance: ["az-ubuntu-2004", "az-ubuntu-2204"]
+        runtimeclass:
+          - "kata-qemu"
+          - "kata-clh"
+        instance:
+          - "az-ubuntu-2004"
+          - "az-ubuntu-2204"
+          - "s390x"
+        exclude:
+          - runtimeclass: "kata-clh"
+            instance: "s390x"
     runs-on: ${{ matrix.instance }}
     steps:
+      - name: Take a pre-action for self-hosted runner
+        run: |
+          if [ -f ${HOME}/script/pre_action.sh ]; then
+            ${HOME}/script/pre_action.sh cc-operator
+          fi
+
       - uses: actions/checkout@v4
+
+      - name: Rebase atop of the latest target branch
+        run: |
+          ./tests/git-helper.sh "rebase-atop-of-the-latest-target-branch"
+        env:
+          TARGET_BRANCH: ${{ inputs.target-branch }}
 
       - name: Install deps
         run: |
@@ -29,4 +52,17 @@ jobs:
         run: |
           cd tests/e2e
           export PATH="$PATH:/usr/local/bin"
-          ./run-local.sh -r "${{ matrix.runtimeclass }}" -u
+          args="-u"
+          if [ $RUNNING_INSTANCE = "s390x" ]; then
+            args=""
+          fi
+          ./run-local.sh -r "${{ matrix.runtimeclass }}" "${args}"
+        env:
+          RUNNING_INSTANCE: ${{ matrix.instance }}
+
+      - name: Take a post-action
+        if: always()
+        run: |
+          if [ -f ${HOME}/script/post_action.sh ]; then
+            ${HOME}/script/post_action.sh cc-operator
+          fi
--- a/tests/git-helper.sh
+++ b/tests/git-helper.sh
@@ -1,0 +1,63 @@
+#!/usr/bin/env bash
+#
+# Copyright Confidential Containers Contributors
+#
+# SPDX-License-Identifier: Apache-2.0
+#
+
+set -o errexit
+set -o nounset
+set -o pipefail
+
+function install_jq() {
+	if [ -n "$(command -v apt-get)" ]; then
+		sudo apt-get update
+		sudo apt-get install -y jq
+	elif [ -n "$(command -v yum)" ]; then
+		sudo yum install -y epel-release
+		sudo yum install -y jq
+	else
+		>&2 echo "No supported package manager found"
+		exit 1
+	fi
+}
+
+function configure_github() {
+	if [ ! command -v jq &> /dev/null ]; then
+		echo "jq is not installed, installing it"
+		install_jq
+	fi
+	USERNAME=$(jq -r '.pull_request.user.login' "$GITHUB_EVENT_PATH")
+	EMAIL=$(jq -r '.pull_request.user.email' "$GITHUB_EVENT_PATH")
+	# if the email is null, stuff with a dummy email
+	if [ "${EMAIL}" == "null" ]; then
+		EMAIL="dummy@email.address"
+	fi
+	echo "Adding user name ${USERNAME} and email ${EMAIL} to the local git repo"
+	git config user.name "${USERNAME}"
+	git config user.email "${EMAIL}"
+}
+
+function rebase_atop_of_the_latest_target_branch() {
+	if [ -n "${TARGET_BRANCH}" ]; then
+		configure_github
+		echo "Rebasing atop of the latest ${TARGET_BRANCH}"
+		# Recover from any previous rebase left halfway
+		git rebase --abort 2> /dev/null || true
+		if ! git rebase origin/${TARGET_BRANCH}; then
+			>&2 echo "Rebase failed, exiting"
+			exit 1
+		fi
+	fi
+}
+
+function main() {
+	action="${1:-}"
+
+	case "${action}" in
+	rebase-atop-of-the-latest-target-branch) rebase_atop_of_the_latest_target_branch;;
+	*) >&2 echo "Invalid argument"; exit 2 ;;
+	esac
+}
+
+main "$@"
--- a/.github/workflows/ccruntime-nightly.yaml
+++ b/.github/workflows/ccruntime-nightly.yaml
@@ -1,0 +1,24 @@
+name: ccruntime e2e test nightly
+on:
+  schedule:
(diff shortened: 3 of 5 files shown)


BbolroC:
> Hey @BbolroC, looks good to me. Regarding the pre- and post-action scripts and `if: ${{ matrix.arch == 's390x' }}` vs. `if: always()`: You're already using `runs-on: s390x` and a self-hosted runner for this workflow, so does that not solve the problem?

Hi, thank you for your comment. I'd like to add more information to further clarify the conflict I previously explained:

Let's say we would like to incorporate 2 different architectures by using 2 runners named `x86_64` and `s390x`, respectively.

```yaml
name: Example Workflow

jobs:
  job1:
    matrix:
      os: [x86_64, s390x]
    runs-on: {{ matrix.os }}

    steps:
      - name: Pre Step
        run: echo "Running Pre Step"
        if: {{ matrix.os == 's390x' }}

      - name: Main Step
        run: echo "Running Main Step"

      - name: Post Step
        if: {{ matrix.os == 's390x' }}
        run: echo "Running Post Step"
```

Thinking that the {pre,post} steps are only required for `s390x`, the conditional `if: {{ matrix.os == 's390x' }}` at the post step prevents the step from achieving what `if: always()` did before because it is not 100% true. Consequently, the post step fails to execute when other steps encounter an error.

Upon further consideration, a potential resolution could involve passing a matrix element as an environment variable into a step, like:

```yaml
      - name: Post Step
        if: always()
        run: echo "Dealing with a step selectively based on an environment variable `OS`"
        env:
          OS: {{ matrix.os }}
```

I can confidently state that the introduction of the {pre,post} actions wouldn't pose any issues for incorporating workflows across multiple architectures. Thanks.

.github/workflows/ci-s390x.yaml
```diff
@@ -0,0 +1,36 @@
...
+    runs-on: s390x
+
+    strategy:
+      matrix:
+        runtimeclass: ["kata-qemu"]
+
+    steps:
+      - name: Adjust a permission for repo
+        run: |
+          sudo chown -R $USER:$USER $GITHUB_WORKSPACE
```
wainersm, line 19: hi @BbolroC should this chown call be part of the pre-action too?

wainersm:
@ryansavino could you also jump in this discussion? Essentially @BbolroC, @portersrc and I have figured out ways to have a multi-arch and non-TEE/TEE integrated workflow for running the e2e tests. I do believe that for SEV/SNP you will need to have pre/post actions as well.

.github/workflows/ci-s390x.yaml
BbolroC: Yeah, you are totally right. Thanks for pointing it out. I will update the PR.

FYI: this optimisation will also apply to kata GHA (https://github.com/kata-containers/kata-containers/pull/8649) soon. 😉

BbolroC:
As discussed above, the workflows introduced in this PR are incorporated into the existing x86_64 workflow `ccruntime_e2e.yaml`. So the workflows of interest are triggered not only on PR, but also nightly.

@wainersm @stevenhorsman @fidencio  please take a look at this and give me a feedback. Thanks!

.github/workflows/ccruntime-on-push.yaml
```diff
@@ -0,0 +1,17 @@
+name: ccruntime e2e test on-push
+on:
+  pull_request_target:
```
wainersm, line 3: Changing from `pull_request` to `pull_request_target` will change the behavior of ` actions/checkout@v4`, i.e., it will check out the pull request tree which is not rebased with main. We will need to rebase the check'ed code ourselves to emulate the same behavior of `pull_request`. This is a well-known problem, we can fix it as kata CI and in a follow up PR if you prefer.

.github/workflows/ccruntime-on-push.yaml
```diff
@@ -0,0 +1,17 @@
+name: ccruntime e2e test on-push
```
wainersm, line 1: Nit: not use `on-push` on the workflow name, file and jobs as this workflow is triggered on pull request.

ldoktor (commented):
I think @wainersm said it all, it looks almost ready, just the little naming issue and (if you have time) the `pull_request_target` rebase it should be ready to go.

.github/workflows/ccruntime_e2e.yaml
```diff
@@ -29,4 +41,17 @@ jobs:
         run: |
           cd tests/e2e
           export PATH="$PATH:/usr/local/bin"
-          ./run-local.sh -r "${{ matrix.runtimeclass }}" -u
+          args="-u"
+          if [ $RUNNING_INSTANCE = "s390x" ]; then
+            args=""
```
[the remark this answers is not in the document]
BbolroC: Yes, it was concluded that the option `-u` is not suitable for a self-hosted runner context. The functionality of the option is delegated to the {pre,post}-action.

.github/workflows/ccruntime-on-push.yaml
BbolroC: Yeah, then I will borrow the script and place it in this PR rather than instituting it in the next PR.

.github/workflows/ccruntime-on-push.yaml
BbolroC: Yes, I will rename ones relevant to this comment. Thanks for the comment.

BbolroC:
@wainersm @ldoktor I've updated the PR based on your feedback. Looking forward to the next step. Thanks!

tests/git-helper.sh
```diff
@@ -0,0 +1,63 @@
...
+}
+
+function rebase_atop_of_the_latest_target_branch() {
+	if [ -n "${TARGET_BRANCH}" ]; then
+		configure_github
+		echo "Rebasing atop of the latest ${TARGET_BRANCH}"
+		# Recover from any previous rebase left halfway
+		git rebase --abort 2> /dev/null || true
+		if ! git rebase origin/${TARGET_BRANCH}; then
+			>&2 echo "Rebase failed, exiting"
```
stevenhorsman, line 48: Hey Choi in the kata-containers version there was a different path for self-hosted runner than deleted the workspace directory to avoid merge conflicts impacting runs IIRC - is that needed here too?

stevenhorsman: Great, thanks for the explanation!

stevenhorsman (approved):
LGTM

BbolroC:
Now, we are facing the `pull_request_target` problem. Please help me out!

stevenhorsman:
> Now, we are facing the `pull_request_target` problem (the old checks marked as required are not triggered any more). Please help me out!

Just to confirm, this is just a one off where we need to remove them as required and then after this merge we've have new jobs to make required that superceded them?

BbolroC:
> > Now, we are facing the `pull_request_target` problem (the old checks marked as required are not triggered any more). Please help me out!
> 
> Just to confirm, this is just a one off where we need to remove them as required and then after this merge we've have new jobs to make required that superceded them?

Yes, I think so, unfortunately. If we don't like this approach, let's stop proceeding things and start re-discussing what could be a better way to trigger the tests on the expected github events for a PR in a protective way.

stevenhorsman:
> Yes, I think so, unfortunately. If we don't like this approach, let's stop proceeding things and start re-discussing what could be a better way to trigger the tests on the expected github events for a PR in a protective (resource) way.

I think that makes sense to take the hit now, especially given we are far away from releases. @wainersm are you happy with this?

ldoktor (approved):
Thanks for the changes, what is the problem with the required tests? I remember the discussion but still new to this repo. In different repos when GH CI did not scale we simply tested the PR locally and pushed things via cmdline (where the required conditions are not checked). Worst case we can always revert if the GH runners find the main version broken.

wainersm:
> > Yes, I think so, unfortunately. If we don't like this approach, let's stop proceeding things and start re-discussing what could be a better way to trigger the tests on the expected github events for a PR in a protective (resource) way.
> 
> I think that makes sense to take the hit now, especially given we are far away from releases. @wainersm are you happy with this?

Yes @stevenhorsman , I'm happy with that. 

@BbolroC I added an entry on the agenda of our meeting today to discuss the methods we have used to protect the self-hosted runners.

wainersm:
> Thanks for the changes, what is the problem with the required tests? I remember the discussion but still new to this repo. In different repos when GH CI did not scale we simply tested the PR locally and pushed things via cmdline (where the required conditions are not checked). Worst case we can always revert if the GH runners find the main version broken.

Hi @ldoktor , let me see if I can explain accurately...

This PR change the current jobs so that they are triggered on a different event (`pull_request_target`) and now requires the `ok-to-test` label. If we add the `ok-to-test` label on this PR the new jobs will be triggered but will fail because on a `pull_request_target` it reads the workflows from the target branch (`main`) which doesn't exist until this PR is merged. I.e. it is chicken-and-eggs situation. So we will need to make the current jobs non-required so that we can merge the PR first.

PR #303, opened 2023-12-13
controller: Handle snapshotter per handler

fidencio:
Let's start properly setting a specific snapshotter per runtime handler configured for containerd.

This work depends on a work done on the Kata Containers side to better support setting snapshotters per runtime handler.

The PR from the Kata Containers side is:
https://github.com/kata-containers/kata-containers/pull/8655.

### DIFF
--- a/controllers/ccruntime_controller.go
+++ b/controllers/ccruntime_controller.go
@@ -628,20 +628,19 @@ func (r *CcRuntimeReconciler) processDaemonset(operation DaemonOperation) *appsv
 		createDefaultRuntimeClass = "true"
 	}
 
-	var runtimeClassNames []string
-	var snapshotter = ""
+	var shims []string
+	var snapshotter_handler_mapping []string
 	for _, runtimeClass := range r.ccRuntime.Spec.Config.RuntimeClasses {
-		runtimeClassNames = append(runtimeClassNames, runtimeClass.Name)
-		// FIXME: This will need to be changed by the moment the kata-containers
-		//        payload script supports setting one snapshotter per runtime handler.
-		//        For now, for the v0.8.0 release, we're fine assuming that all the
-		//        set snapshotters are going to be the same.
-		if snapshotter == "" && runtimeClass.Snapshotter != "" {
-			snapshotter = runtimeClass.Snapshotter
+		// Similarly to what's being done for the default shim, let's remove
+		// the "kata-" prefix from the runtime class names
+		shim := strings.TrimPrefix(runtimeClass.Name, "kata-")
+		shims = append(shims, shim)
+
+		if runtimeClass.Snapshotter != "" {
+			mapping := shim + ":" + runtimeClass.Snapshotter
+			snapshotter_handler_mapping = append(snapshotter_handler_mapping, mapping)
 		}
 	}
-	var runtimeClasses = strings.Join(runtimeClassNames, " ")
-	var shims = strings.ReplaceAll(runtimeClasses, "kata-", "")
 
 	var envVars = []corev1.EnvVar{
 		{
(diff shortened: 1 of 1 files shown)


controllers/ccruntime_controller.go
```diff
@@ -629,15 +629,16 @@ func (r *CcRuntimeReconciler) processDaemonset(operation DaemonOperation) *appsv
...
 	for _, runtimeClass := range r.ccRuntime.Spec.Config.RuntimeClasses {
 		runtimeClassNames = append(runtimeClassNames, runtimeClass.Name)
-		// FIXME: This will need to be changed by the moment the kata-containers
-		//        payload script supports setting one snapshotter per runtime handler.
-		//        For now, for the v0.8.0 release, we're fine assuming that all the
-		//        set snapshotters are going to be the same.
-		if snapshotter == "" && runtimeClass.Snapshotter != "" {
-			snapshotter = runtimeClass.Snapshotter
+
+		if runtimeClass.Snapshotter != "" {
```
wainersm, line 636: I think that here it should be `== ""`.

One alternative is to use Go map, convert to json then to string:
```
snapshotters := map[string]string{"kata-qemu": "nydus", "kata-clh": "nydus"}
j, _ := json.Marshal(snapshotters)
snapshottersStr := string(j)
```

But then `SNAPSHOTTER_HANDLER_MAPPING` should be processed as a json string on kata side (i.e. `{"k1":"v1","k2":"v2"}` vs  `"k1":"v1","k2":"v2"`). One advantage of passing a json string is that in the future we can pass properties alongside if needed.

stevenhorsman: Wainer is correct - the current logic produces:
`Validating the snapshotter-handler mapping: ",kata-clh:nyduskata-qemu:nyduskata-qemu-tdx:nyduskata-qemu-sev:nyduskata-qemu-snp:nydus"`. I'll try and fix it, rebase and push to this branch

stevenhorsman: I managed to force push in the end to hopefully fix this.

stevenhorsman:
During testing of this I found a bug in kata-deploy, so waiting on https://github.com/kata-containers/kata-containers/pull/8721 to be merged to fix this.

fidencio:
I found a few more issues on the kata-deploy side, all addressed as part of: https://github.com/kata-containers/kata-containers/pull/8733

I've also created a new PR on the Operator side in order to test with the kata-deploy payload, before this one / and the Kata Containers one gets merged: https://github.com/confidential-containers/operator/pull/310

stevenhorsman (approved):
LGTM now. Thanks!

beraldoleal (approved):
LGTM, thanks.

controllers/ccruntime_controller.go
```diff
@@ -629,15 +629,17 @@ func (r *CcRuntimeReconciler) processDaemonset(operation DaemonOperation) *appsv
...
+			shim := strings.TrimPrefix(runtimeClass.Name, "kata-")
+			if snapshotter_handler_mapping == "" {
+				snapshotter_handler_mapping += shim + ":" + runtimeClass.Snapshotter
+			} else {
+				snapshotter_handler_mapping += "," + shim + ":" + runtimeClass.Snapshotter
+			}
```
mythi, line 642: this `kata-` prefix processing is repeated below. Can we do it only once? I also think preparing `snapshotter_handler_mapping` can follow the same `strings.Join()` approach used by `runtimeClassNames`.

fidencio: So, the prefix processing is done for a different reason here, and looking at the code is quite hard to make this only once, as kata-containers explicitly requires the shim name for all the operations, and for the sake of the user's experience the operator requires a full runtime class name (as in kata-qemu).

About the string.Join(), this can be done, I just need to either make sure `, ` is correctly treated on the kata-containers side, or that we remove the last `,` that would be added by the strings.Join().

fidencio: Cool, I've updated the PR and simplified a little bit the logic.
Thanks for the rveiew, @mythi!

mythi (approved):
LGTM

PR #308, opened 2023-12-22
Controller properly set snapshotter per runtime handler

stevenhorsman:
Temp PR to rebase Fabiano's branch and see if Wainer's suggestion fixes the test issues


stevenhorsman:
I've pushed the fix to Fabiano's PR, so this can close now

PR #313, opened 2024-01-08
cli tool to change containerd config

mkulke:


fidencio:
@mkulke, one thing that's been crossing my mind is whether we need a new tool for this or whether we'd be good enough relying on `tomlq` for this as we did on the Kata Containers side.

I'd like to know your opinion on this,

mkulke:
> @mkulke, one thing that's been crossing my mind is whether we need a new tool for this or whether we'd be good enough relying on `tomlq` for this as we did on the Kata Containers side.
> 
> I'd like to know your opinion on this,

as said above, it doesn't matter whether you use a library or a cli tool, but if you want to have some confidence in what is happening, i.e. write tests and handle all those potential errors properly, this is probably easier to do in a dedicated tool.

a simple tomlq invocation will just handle the happy path, the rest you have to code (or opt to ignore) in bash

fidencio:
> a simple tomlq invocation will just handle the happy path, the rest you have to code (or opt to ignore) in bash

Fair enough! :-)

fidencio:
@mkulke: https://github.com/containerd/nydus-snapshotter/pull/570

Issue #320, opened 2024-01-16
kustomize does not allow 2nd level override


BbolroC:
**Describe the bug**
Unlike `default` or `peer-pods`, a kustomize manifest for `s390x` overrides the resources on `default` rather than `base`, which is the 2nd level of override. This is not allowed in kustomize and it is verified that it does not work to rename an image or a tag for the architecture. The override for `s390x` should be made based on `base` like others.

PR #321, opened 2024-01-16
kata-payload: Make resources for s390x overridden by base

BbolroC:
Due to the fact that kustomize does not allow the 2nd level of resource patch, this commit is to make the resources overridden by base directly.

Fixes: #320


stevenhorsman (approved):
Thanks for the fix @BbolroC

fitzthum (approved):
LGTM

PR #348, opened 2024-02-05
CI: Use ubuntu-22.04 and not ubuntu-latest in workflows

portersrc:
We should at least be consistent with the ubuntu base images we're running in the workflows. I also opt to use ubuntu-22.04 and not ubuntu-latest.

Other repos in the org can take a similar change if reviewers want.

### DIFF
--- a/.github/workflows/docker-publish-latest-on-merge.yml
+++ b/.github/workflows/docker-publish-latest-on-merge.yml
@@ -22,7 +22,7 @@ env:
 jobs:
   build:
 
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-22.04
     permissions:
       contents: read
       packages: write
--- a/.github/workflows/docker-publish-on-tag.yml
+++ b/.github/workflows/docker-publish-on-tag.yml
@@ -22,7 +22,7 @@ env:
 jobs:
   build:
 
-    runs-on: ubuntu-latest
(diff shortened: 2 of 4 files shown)


ldoktor (approved):
Yes, better to be explicit. The 22.04 is currently the latest so nothing actually changes but in the future we'd be surprised.

stevenhorsman (approved):
Makes sense to me. Thanks Chris!

wainersm:
hi @BbolroC !

The s390x post-action step failed: https://github.com/confidential-containers/operator/actions/runs/7796965300/job/21262642550?pr=348

I wondering if there is something we can do to make that step more reliable.

stevenhorsman:
> hi @BbolroC !
> 
> The s390x post-action step failed: https://github.com/confidential-containers/operator/actions/runs/7796965300/job/21262642550?pr=348
> 
> I wondering if there is something we can do to make that step more reliable.

Choi might be able to confirm, but he updated and re-registered the s390x runners earlier, which might have coincided with that run?

BbolroC:
@wainersm @stevenhorsman yeah, the time matches. I was in the middle of the update at the time.

BbolroC:
The test for s390x has been successfully passed. Sorry for the inconvenience.

GabyCT:
lgtm

PR #355, opened 2024-03-20
crd: Expose agent{Https,No}Proxy

fidencio:
Depending on the environment where we deploy Confidential Containers, setting up the proxy is required for the agent to be able to connect with the external world.

With that in mind, mainly considering this is needed for the basic TDX CI, let's ensure we expose to the users a way to set it up.

### DIFF
--- a/config/samples/ccruntime/default/kustomization.yaml
+++ b/config/samples/ccruntime/default/kustomization.yaml
@@ -46,5 +46,15 @@ patches:
         # the case, please, set this variable to `true`
         - name: "INSTALL_OFFICIAL_CONTAINERD"
           value: "false"
+        # If set, the Kata Containers agent https_proxy will be set to the
+        # specified value allowing then the pod sandbox to correctly pull
+        # images in such environment. 
+        # - name: "AGENT_HTTPS_PROXY"
+        #   value: "https://proxy.example.com:3129"
+        # If set, the Kata Containers agent no_proxy will be set to the
+        # specified value allowing then the pod sandbox to correctly pull
+        # images in such environment.
+        # - name: "AGENT_NO_PROXY"
+        #   value: "*.test.example.com,.example.org,127.0.0.0/8"
   target:
     kind: CcRuntime


ChengyuZhu6 (approved):
LGTM thanks @fidencio !

api/v1beta1/ccruntime_types.go
```diff
@@ -168,6 +168,16 @@ type CcInstallConfig struct {
...
+	// This specifies the HTTPS_PROXY that may be required to be set for the Kata Containers agent
+	// when performing the image pull inside the guest (either using nydus snapshotter with containerd
+	// or CRI-O)
+	AgentHttpsProxy string `json:"agentHttpsProxy,omitempty"`
+
+	// This specifies the NO_PROXY that may be required to be set for the Kata Containers agent
+	// when performing the image pull inside the guest (either using nydus snapshotter with containerd
+	// or CRI-O)
+	AgentNoProxy string `json:"agentNoProxy,omitempty"`
+
```
bpradipt, line 180: Since this is kata runtime specific (and I'm assuming doesn't apply to enclave-cc), how about putting these in a separate Kata agent specific struct so that the code is easier to read and update if new agent config params gets added ?

fidencio: That's a very good point, @bpradipt, let me revisit this one soon.

mythi: Is the same not possible with `EnvironmentVariables`? Do you have the code somewhere that takes these into the kata agent?

wainersm: Based on ^^^ I added a do-not-merge label.

fidencio:
I've taken @mythi's suggestion and instead of exposing it, we're just leaving a hint to the user, in the default kustomization, of what they can uncomment in order to have proxies working as expected.

mythi (approved):
LGTM. I prefer this approach over adding proxies to the "API"

bpradipt (approved):
/lgtm

PR #362, opened 2024-04-23
update peerpod and peerpodconfig deps

bpradipt:
This switches the deps to the latest code which
includes a fix to handle the CAA daemonset spec updates


stevenhorsman (approved):
LGTM and matches the version we merged in https://github.com/confidential-containers/cloud-api-adaptor/pull/1814. Thanks!

go.mod
```diff
@@ -3,9 +3,9 @@ module github.com/confidential-containers/operator
 go 1.21
 
 require (
-	github.com/confidential-containers/cloud-api-adaptor/peerpod-ctrl v0.8.0-alpha.1.0.20240204080936-372e4e72f7fe
-	github.com/confidential-containers/cloud-api-adaptor/peerpodconfig-ctrl v0.8.0-alpha.1.0.20240204080936-372e4e72f7fe
-	github.com/go-logr/logr v1.2.4
+	github.com/confidential-containers/cloud-api-adaptor/src/peerpod-ctrl v0.8.3-0.20240422235106-567287db9814
+	github.com/confidential-containers/cloud-api-adaptor/src/peerpodconfig-ctrl v0.8.3-0.20240422235106-567287db9814
```
mythi, line 7: is it not possible to use `v0.8.2` here?

mythi: Just curious if it was possible to just make v0.8.3 with the fix? I'm fine with this too however.

bpradipt: @mythi unfortunately no as the bug fix went post 0.8.2
I think before 0.9.0 we'll need to create another release for cloud-api-adaptor due to the switch to main and update it here with the release version

bpradipt: For some (most likely network) reasons my comment came very late..

huoqifeng (approved):
LGTM

go.mod
mythi: @bpradipt yes but what prevents us making a v0.8.3 if the fix is important?

bpradipt: @mythi It's the time required to make a release. I'll let @stevenhorsman comment on the feasibility of cutting a quick 0.8.3 release.
Also @mythi may I ask why you insist on a release and not a commit during the development phase ?

mythi: To make it clear, I don't insist on a release: 

> Just curious if it was possible to just make v0.8.3 with the fix? I'm fine with this too however.

bpradipt: Ah ok, I thought there is some dependency or something else missing..
Let's wait to hear from @stevenhorsman on 0.8.3 release feasibility

stevenhorsman: I think it depends on if we want a full CAA release, or just tag the two go modules being used here. I've added it to the agenda of today's community meeting to discuss.

stevenhorsman: @mythi, hey Mikko, just to summarise what we discussed on our call:
We don't want the overhead of doing a full peer-pods release, just to do an update to a component that we aren't directly consuming get, so our remaining options are:
- Tag the two do modules involved with 0.8.3. This is pretty easy, but means that if we do a subsequent release in the 0.8.x lineage we'll need to skip one and go to 0.8.4, but that's not a big cost IMHO (I'm not sure if everyone agrees with me there)
- We stick with the commit reference as done in this PR, which I understand it's ideal as this would be unlikely to change before an operator 0.9 release, but it's not on the direct user path for 0.9 as we won't be instructing anyone to install the peer-pods version of the CoCo controller manager in this release.

I'm not sure if you have a strong opinion either way on those two options? Thanks!

mythi: > * We stick with the commit reference as done in this PR, which I understand it's ideal as this would be unlikely to change before an operator 0.9 release, but it's not on the direct user path for 0.9 as we won't be instructing anyone to install the peer-pods version of the CoCo controller manager in this release.

This is totally fine, thanks for the follow-up! I think going forward, it would be good to make sure tagged versions can be used here and let, e.g., dependabot to take care of PRs like this if the goal is to be able to use CoCo operator also for the PeerPod functionality.

mythi (approved):

go.mod
stevenhorsman: Agreed - as we discussed in the 0.9 checklist PR I think we are going to move to releasing peer pods before the operator and that should avoid problems like this, unfortunately for reasons peer pods isn't part of the 0.9, so it's a bit different this time.

stevenhorsman:
I guess now that 0.9-alpha0 has released we can merge this

PR #373, opened 2024-05-02
tests: remove deprecated ansible option

fitzthum:
Ansible used to have a "feature" where it would warn you if you wrote a shell command that could be implemented with a module instead. The shell option had an option to disable these warnings.

These features were deprecated in ansible 2.11 and removed in anbsible 2.14. This PR removes the warn option. Otherwise the playbooks won't work. 

This does mean that people using an older version of ansible might see the warnings.


ldoktor (approved):
Yes indeed, thanks for fixing it

GabyCT:
lgtm

wainersm (approved):
Thanks @fitzthum !

PR #375, opened 2024-05-03
ci: update enclave-cc tests

mythi:
Now that both e2e (for PRs) and cicd (nightly) tests are using the updated payloads with the new Occlum instance path, we can drop the sed lines and update the test deployment use the new path directly.

### DIFF
--- a/.github/workflows/enclave-cc-cicd.yaml
+++ b/.github/workflows/enclave-cc-cicd.yaml
@@ -55,6 +55,5 @@ jobs:
 
       - name: Deploy sample workload
         run: |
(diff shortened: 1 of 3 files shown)


ldoktor (approved):
Yep, those are the only 2 places using it and they both are patching it.

portersrc:
thanks @mythi, looks good to me

portersrc (approved):

portersrc:
[Current failing test](https://github.com/confidential-containers/operator/actions/runs/8936491573/job/24546929649?pr=375#step:6:1406) may be network-related
``` 
The connection to the server 10.10.1.4:6443 was refused - did you specify the right host or port?
Error:  runtimeclass kata-qemu is not up
```

fitzthum (approved):
LGTM

PR #377, opened 2024-05-03
Go back to latest payloads after release

fitzthum:
The release is done, so let's go back to the upstream bundles for enclave-cc, kata, and reqs-deploy

### DIFF
--- a/config/samples/ccruntime/base/ccruntime.yaml
+++ b/config/samples/ccruntime/base/ccruntime.yaml
@@ -24,6 +24,8 @@ spec:
         name: kata-artifacts
       - mountPath: /usr/local/bin/
         name: local-bin
+      - mountPath: /host/
+        name: host
     installerVolumes:
       - hostPath:
           path: /etc/crio/
@@ -41,6 +43,10 @@ spec:
           path: /usr/local/bin/
           type: ""
         name: local-bin
+      - hostPath:
+          path: /
+          type: ""
+        name: host
     installCmd: ["/opt/kata-artifacts/scripts/kata-deploy.sh", "install"]
     uninstallCmd: ["/opt/kata-artifacts/scripts/kata-deploy.sh", "cleanup"]
     cleanupCmd: ["/opt/kata-artifacts/scripts/kata-deploy.sh", "reset"]
--- a/config/samples/ccruntime/default/kustomization.yaml
+++ b/config/samples/ccruntime/default/kustomization.yaml
@@ -8,9 +8,10 @@ resources:
 
 images:
 - name: quay.io/confidential-containers/reqs-payload
-  newTag: 290ae5f8432f81754dcdc5708ed1488e3986d79b
+  newTag: latest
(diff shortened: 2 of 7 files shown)


ldoktor (commented):
Hello @fitzthum, wouldn't it be clearer to use `git revert` as you are effectively reverting the 9d14530afece97840799a5b0e2fb8e18fd2a1056 f36b294d2d7b701505ae9f00b2f5a9c833aa0019 and actually even aa2706048c0fcd29c0191627de82c8b0f1b51118 ? I think doing so would be a nice practice in the future.

fitzthum:
> Hello @fitzthum, wouldn't it be clearer to use git revert as you are effectively reverting the https://github.com/confidential-containers/operator/commit/9d14530afece97840799a5b0e2fb8e18fd2a1056 https://github.com/confidential-containers/operator/commit/f36b294d2d7b701505ae9f00b2f5a9c833aa0019 and actually even https://github.com/confidential-containers/operator/commit/aa2706048c0fcd29c0191627de82c8b0f1b51118 ? I think doing so would be a nice practice in the future.

I was thinking about this, but it seemed weird to revert the fairly old change for enclave-cc (we forgot to update those on the last release). I don't really have a preference either way. You might want to comment on https://github.com/confidential-containers/confidential-containers/pull/210 where we are fixing up the release checklist for next time. I think we will probably start using a script to make these changes in the future as well.

wainersm (approved):
I also liked the `git revert` idea. Maybe next release we can update the files so that it's git revert-able.

fidencio (changes requested):
@fitzthum, the PR itself is not wrong, but it'll need a follow-up commit (as part of the very same PR).

Once we revert to use the latest from kata-containers, Operator jobs will break, as on the Kata Containers side a change was made but not reflected here, and it's my fault, see:
* https://github.com/kata-containers/kata-containers/blob/9a6d8d83308870ca662b1b811b1bb7c42b96cfba/tools/packaging/kata-deploy/kata-deploy/base/kata-deploy.yaml#L61-L62
* https://github.com/kata-containers/kata-containers/blob/9a6d8d83308870ca662b1b811b1bb7c42b96cfba/tools/packaging/kata-deploy/kata-deploy/base/kata-deploy.yaml#L77-L79

That mount has to be propagated here:
* https://github.com/confidential-containers/operator/blob/0e8b344e01e85b09684baee0392aa553395f24a7/config/samples/ccruntime/base/ccruntime.yaml#L18-L43

Thanks, and let me know if you face any issue updating the file.

fitzthum:
Added the host mount.

I don't want this PR to become a catch-all for changes we pick up from Kata. Ideally it should be merged immediately after the release.

fidencio:
> I don't want this PR to become a catch-all for changes we pick up from Kata. Ideally it should be merged immediately after the release.

I agree, but if the tests were not passing, we fix the minimum enough to get the CI green (which is the case at this point).

fidencio (approved):
lgtm, thanks @fitzthum!

PR #379, opened 2024-05-15
pre-reqs: Bump reqs-payload image

stevenhorsman:
- After we updated to pick up nydus-snapshotter v0.3.13 in the pre-reqs image, start picking up this version.

### DIFF
--- a/config/samples/ccruntime/default/kustomization.yaml
+++ b/config/samples/ccruntime/default/kustomization.yaml
@@ -8,7 +8,7 @@ resources:
 
 images:
 - name: quay.io/confidential-containers/reqs-payload
-  newTag: e36bf9d3214421bad4f9e8e054b44c7008d33ec7
+  newTag: 290ae5f8432f81754dcdc5708ed1488e3986d79b
 - name: quay.io/kata-containers/kata-deploy
   newTag: 3.4.0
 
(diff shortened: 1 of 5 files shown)


fitzthum (approved):
LGTM

ldoktor (approved):
This does cover all of the places although the commit message is a bit cryptical (maybe because I'm not native). Something like `pre-reqs image had been recently updated to use nydus-snapshotter v0.3.13, use that version in our sample configs.` would be IMO clearer, what do you think? (again, I'm not native)

stevenhorsman:
> This does cover all of the places although the commit message is a bit cryptical (maybe because I'm not native). Something like `pre-reqs image had been recently updated to use nydus-snapshotter v0.3.13, use that version in our sample configs.` would be IMO clearer, what do you think? (again, I'm not native)

That's fair - Wainer recently accused my commit messages of sounding like they were AI generated, so I need to work on that!

stevenhorsman:
> This does cover all of the places although the commit message is a bit cryptical (maybe because I'm not native). Something like `pre-reqs image had been recently updated to use nydus-snapshotter v0.3.13, use that version in our sample configs.` would be IMO clearer, what do you think? (again, I'm not native)

@ldoktor - I hope this update is clearer? If not - let me know!

PR #385, opened 2024-06-11
Makefile: scripts: Add build args for proxy when using docker build

GabyCT:
This PR adds build args for proxy when using docker build, this is specially needed when we are behind a proxy to avoid failures.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -52,6 +52,9 @@ IMG ?= controller:latest
 # ENVTEST_K8S_VERSION refers to the version of kubebuilder assets to be downloaded by envtest binary.
 ENVTEST_K8S_VERSION = 1.24.2
 
+http_proxy := ""
+https_proxy := ""
+
 # Get the currently used golang install path (in GOPATH/bin, unless GOBIN is set)
 ifeq (,$(shell go env GOBIN))
 GOBIN=$(shell go env GOPATH)/bin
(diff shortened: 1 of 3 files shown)


fidencio (approved):
lgtm, thanks @GabyCT!

stevenhorsman (commented):
The code LGTM, but one question about where the `http_proxy` and `https_proxy` values are set - is the expectation that these environment variables are set on the self-hosted runner independently of the operator workflow?

GabyCT:
@stevenhorsman changes applied

Makefile
```diff
@@ -143,9 +146,9 @@ run: manifests generate fmt vet ## Run a controller from your host.
 docker-build: test ## Build docker image with the manager.
 ifneq (, $(PEERPODS))
 	@echo PEERPODS is enabled
-	docker build -t ${IMG} -f Dockerfile.peerpods .
+	docker build --build-arg http_proxy=$(http_proxy) --build-arg https_proxy=$(https_proxy) -t ${IMG} -f Dockerfile.peerpods .
```
mythi, line 149: is there a reason why this approach was chosen over the better alternative which just setting the `~/.docker/config.json`?

GabyCT: @mythi I have the proxies at the config but seems that when building the dockerfile is still failing without those arguments

GabyCT:
@stevenhorsman now the CI for tdx is passing

stevenhorsman (approved):
Thanks @GabyCT

Makefile
mythi: I don't prefer this approach. It'd be good to understand *why* the best option does not work

PR #394, opened 2024-07-12
ci: Fix up amd runner labels

stevenhorsman:
- The instance need to be a label on the GHA runner, not the runner name, so fix this


BbolroC (approved):
LGTM, thanks @stevenhorsman !

stevenhorsman:
The amd runners won't trigger because this PR is still runner the `main` versions of the workflow

fitzthum (approved):
LGTM

PR #403, opened 2024-07-31
GHA: Remove {pre,post}-action steps for self-hosted runners

BbolroC:
The following hooks:

- ACTIONS_RUNNER_HOOK_JOB_STARTED
- ACTIONS_RUNNER_HOOK_JOB_COMPLETED

could perfectly replace the existing {pre,post}-action scripts and will make a workflow independent of the runner context. This PR wipes out all GHA steps where the actions are triggered.

@AdithyaKrishnan please let me know when the runner for SEV/SNP is ready for this. I will ask for review when you guys are ready. Thanks.

### DIFF
--- a/.github/workflows/ccruntime_e2e.yaml
+++ b/.github/workflows/ccruntime_e2e.yaml
@@ -44,12 +44,6 @@ jobs:
             instance: "sev-snp"
     runs-on: ${{ matrix.instance }}
     steps:
-      - name: Take a pre-action for self-hosted runner
-        run: |
-          if [ -f ${HOME}/script/pre_action.sh ]; then
-            ${HOME}/script/pre_action.sh cc-operator
-          fi
-
       - uses: actions/checkout@v4
         with:
           ref: ${{ inputs.commit-hash }}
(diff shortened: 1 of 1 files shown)


ldoktor (approved):
Yep, those env variables seems to do exactly that so once the workers are configured it should be fine removing them here (as far as the `{pre,post}_action.sh` does not require some extra knowledge about the PR that wouldn't be available in the hooks but I think that shouldn't be the case, rigth?)

BbolroC:
> Yep, those env variables seems to do exactly that so once the workers are configured it should be fine removing them here (as far as the `{pre,post}_action.sh` does not require some extra knowledge about the PR that wouldn't be available in the hooks but I think that shouldn't be the case, rigth?)

Yes, `{pre,post}_action.sh` does not depend on the workflow context. So it would be fine to simply remove them from the workflow. Thanks.

AdithyaKrishnan:
@BbolroC AMD SEV and SNP Coco runners are ready. Please go ahead

AdithyaKrishnan (approved):
LGTM

fitzthum (approved):
LGTM

PR #425, opened 2024-08-26
enable CodeQL checks

mythi:
This PR is a reference for few improvements in our repositories

1. enable dependabot github-actions updates using pinned versions (dependabot understands pinned versions nowadays). Pinned versions are considered safe since attackers could re-push tags. The move to pinned versions must be done manually.
2. enable CodeQL checks for our Go based code (to be checked: maintainers might have to enable this in the repo settings also)

Ref: https://github.com/confidential-containers/confidential-containers/issues/46

PR #426, opened 2024-08-27
bump to k8s.io v1.30, Go 1.22.0, and controller-runtime v0.18.0

mythi:


mythi:
The context for this is that #425 will start complaining about Go version syntax in `go.mod`. However, that is not the only reason for the upgrade. we are greatly outdated with the dependencies

bpradipt (approved):
/lgtm
Thanks @mythi.  This has been long due.

Makefile
```diff
@@ -208,7 +208,7 @@ OPERATOR_SDK ?= $(LOCALBIN)/operator-sdk
 
 ## Tool Versions
 KUSTOMIZE_VERSION ?= v5.3.0
-CONTROLLER_TOOLS_VERSION ?= v0.15.0
+CONTROLLER_TOOLS_VERSION ?= v0.16.1
 OPERATOR_SDK_VERSION ?= v1.30.0
```
bpradipt, line 212: SDK can be bumped as well to latest - v1.36.1

mythi: OK. I will update some of these after #425 is merged

main.go
```diff
@@ -87,8 +88,7 @@ func main() {
 
 	mgr, err := ctrl.NewManager(ctrl.GetConfigOrDie(), ctrl.Options{
 		Scheme:                 scheme,
-		MetricsBindAddress:     metricsAddr,
-		Port:                   9443,
```
BbolroC, line 91: Just dumb question: Is it removed just because it is supposed to be deprecated? I was wondering if it is okay for removal. Thanks.

mythi: Correct, `Port` deprecated (and moved to `manager.WebhookServer`) but it's also a Webhook specific setting we don't use so can be dropped.

BbolroC (approved):
LGTM as long as the comment below is addressed, thanks @mythi !

> OK. I will update some of these after https://github.com/confidential-containers/operator/pull/425 is merged

mythi:
moved to draft since I got myself into a rabbit hole with the k8s stuff...

mythi:
@BbolroC @bpradipt mind taking another look, I pushed some additional comments.

Makefile
mythi: done

fidencio (approved):
lgtm, thanks @mythi!

PR #437, opened 2024-09-04
ccruntime: Stop mounting /opt/kata

fidencio:
There's no reason to do so since the last changes in the Kata Containers side, where kata-deploy.sh relies solely on the /host mount.


mythi (approved):
LGTM

is 9d54f90 going to be a temporary one?

ldoktor (approved):
Thanks, with the latest kata-containers it seems to be working well.

PR #461, opened 2024-10-25
Fix warning on the Dockerfile

GabyCT:
This PR fixes the warning of 'FromAsCasing: 'as and FROM keywords casing do not match' when building the Dockerfile.


stevenhorsman (approved):
LGTM. Thanks

fitzthum (approved):
Nice

PR #485, opened 2025-01-22
test: Deploy local ccruntime for s390x

BbolroC:
`quay.io/confidential-containers/reqs-payload` has been used for ccruntime for s390x because `config/sample/default/kustomization.yaml` is patched. This is only valid for x86_64, but s390x.
It should be `config/sample/${ccruntime_overlay}/kustomization.yaml`. Let's use a local image for the platform.

### DIFF
--- a/tests/e2e/operator.sh
+++ b/tests/e2e/operator.sh
@@ -124,7 +124,7 @@ install_ccruntime() {
 	local overlay_dir="${ccruntime_overlay_basedir}/${ccruntime_overlay}"
 
 	# Use the built pre-install image
-	kustomization_set_image  "${ccruntime_overlay_basedir}/default" \
+	kustomization_set_image  "${ccruntime_overlay_basedir}/${ccruntime_overlay}" \
 		"quay.io/confidential-containers/reqs-payload" \
 		"${PRE_INSTALL_IMG}"
 


fitzthum (approved):
LGTM

BbolroC:
I've checked the following to ensure that kustomize is installed: https://github.com/confidential-containers/operator/blob/fc492f9ec61e1abeb732f926cc0b0fc81217e383/tests/e2e/ansible/install_test_deps.yaml#L56-L67

I will update the PR to remove `ensure_kustomize()`.

stevenhorsman (approved):
LGTM. Thanks @BbolroC!

fitzthum:
Well the 20.04 test is failing with the disk pressure thing. We tried cherry-picking something onto #483 to fix that, but that PR is still failing and this PR doesn't have the change.

fidencio:
> Well the 20.04 test is failing with the disk pressure thing. We tried cherry-picking something onto #483 to fix that, but that PR is still failing and this PR doesn't have the change.

It also fails on #483 if that makes it better.

BbolroC:
FYI: this patch works on top of the latest 10 commits on #483: https://github.com/BbolroC/cc-operator/actions/runs/12925399294/job/36046275940

BbolroC:
I'm closing this because #483 has cherry-picked this and the CI has passed. Thanks.

PR #488, opened 2025-01-24
bundle: updates for 0.12.0

stevenhorsman:
I submitted the https://github.com/k8s-operatorhub/community-operators/pull/5611 to publish 0.10.2 bundle in operator hub and these are the changes on our side for the publish.

### DIFF
--- a/config/manager/kustomization.yaml
+++ b/config/manager/kustomization.yaml
@@ -13,4 +13,4 @@ kind: Kustomization
 images:
 - name: controller
   newName: quay.io/confidential-containers/operator
-  newTag: v0.10.0
+  newTag: v0.12.0
--- a/Makefile
+++ b/Makefile
@@ -3,7 +3,7 @@
 # To re-generate a bundle for another specific version without changing the standard setup, you can:
 # - use the VERSION as arg of the bundle target (e.g make bundle VERSION=0.0.2)
 # - use environment variables to overwrite this value (e.g export VERSION=0.0.2)
-VERSION ?= 0.10.0
+VERSION ?= 0.12.0
 ARCH = $(shell uname -m)
 
 # CHANNELS define the bundle channels used in the bundle.
--- a/bundle/manifests/cc-operator-controller-manager-metrics-service_v1_service.yaml
+++ b/bundle/manifests/cc-operator-controller-manager-metrics-service_v1_service.yaml
@@ -10,7 +10,7 @@ spec:
   - name: https
     port: 8443
     protocol: TCP
-    targetPort: https
+    targetPort: 8443
   selector:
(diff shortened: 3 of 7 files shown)


config/manager/kustomization.yaml
```diff
@@ -13,4 +13,4 @@ kind: Kustomization
 images:
 - name: controller
   newName: quay.io/confidential-containers/operator
-  newTag: v0.10.0
+  newTag: v0.12.0
```
mythi, line 16: could be also `latest` so that the kustomization in main HEAD always matches the latest image but `make bundle` seems to force a `VERSION` tag...

mythi (approved):
LGTM, thanks!

config/manager/kustomization.yaml
stevenhorsman: Yeah, it looks like in the past I updated this back to latest post-release (https://github.com/confidential-containers/operator/commit/81a89f9fff01d5678043115ed40c43437d7adee6), so we could add this to our post release tasks? Or alternatively update `config/default/kustomization.yaml` to have
```
images:
- name: quay.io/confidential-containers/operator
  newTag: latest 
```
as I mentioned on slack, which I think should have the same effect?

mythi: > Or alternatively update `config/default/kustomization.yaml` to have

+1

wainersm: Updating config/default/kustomization.yaml to use latest image is the best option IMO.

wainersm (approved):
Thanks for the quick fix @stevenhorsman !

config/manager/kustomization.yaml
stevenhorsman: Cool - we can do that once the release is all tied up I guess

stevenhorsman:
> We do not need a new tag for this release, right?

No, I've got a workaround on the CAA side and I think this should resolve the issue anyway

> Can we make sure to add this to the release checklist? I guess it is part of step 8 but we should maybe clarify that.

Sorry, I'm not sure what you mean by this?

fitzthum:
> Sorry, I'm not sure what you mean by this?

Is there something we have forgotten to do in previous releases? If so we should add it to the release template section.

I missed most of the discussion on this one due to being asleep.

fitzthum (approved):
LGTM

config/manager/kustomization.yaml
mythi: I took a quick look and it seems the overall flow needs rework: `make bundle` sets the image tag but it has no effect since `config/manifest` actually uses `config/release` as the target that sets an image tag that is decoupled from `VERSION` used by `make bundle`.

PR #487, opened 2025-01-23
Release: Update payloads for v0.12.0

cclaudio:
./hack/release-helper.sh update -o v0.12.0 -e v0.11.0 -k 3.13.0

PR #489, opened 2025-01-28
release: Post 0.12.0 release reverts

stevenhorsman:
release: Post-release reverts

This reverts the hard-coding of payload versions
for the 0.12 release done in #487

### DIFF
--- a/config/samples/ccruntime/default/kustomization.yaml
+++ b/config/samples/ccruntime/default/kustomization.yaml
@@ -8,9 +8,10 @@ resources:
 
 images:
 - name: quay.io/confidential-containers/reqs-payload
-  newTag: b495f450fbcd5c199b8635a7dfefe1b5d763c378
+  newTag: latest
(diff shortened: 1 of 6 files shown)


fitzthum (approved):
LGTM

Thanks

ldoktor (approved):
LGTM although it took me a while to find out why the revert commits contain a different hash than the ones in https://github.com/confidential-containers/operator/pull/487 . @fitzthum is there a specific reason not to use merge commit to merge PRs or is it just personal matter of taste?

fitzthum:
>  or is it just personal matter of taste?

Yes. I don't really like merge commits in the history and it can have confusing behavior in some cases. For instance during the release process we check that the latest changes to the pre-reqs dir match the pre-reqs image that we are pulling, but the image ends up being built from the merge commit which you won't see if you go to the directory. I don't really care that much tho tbh.

PR #490, opened 2025-01-30
operator: Use latest operator image

stevenhorsman:
Use the latest operator image in the default config as the base manage kustomization typically has the release tag and the release kustomization also has a release tag pinned, so we want to use default for
development and testing the latest operator after discussion with @mythi @fitzthum & @wainersm

### DIFF
--- a/config/default/kustomization.yaml
+++ b/config/default/kustomization.yaml
@@ -28,6 +28,12 @@ resources:
 # [METRICS] Expose the controller manager metrics service.
 - metrics_service.yaml
 
+# The release config and manage kustomization points to the pinned release, so override this
+# dev config to latest for up to date testing
+images:
+- name: quay.io/confidential-containers/operator
+  newTag: latest
+
 patches:
 # [METRICS] The following patch will enable the metrics endpoint using HTTPS and the port :8443.
 # More info: https://book.kubebuilder.io/reference/metrics


mythi:
In https://github.com/confidential-containers/operator/pull/488#discussion_r1930071616 I made the comment that the bundle build is a bit confusing. I remember we agreed the approach in this PR but the alternative would be to just make [the bundle creation](https://github.com/confidential-containers/operator/blob/4b6b25f42024efa723b2838bb3f583cd033cbdf2/Makefile#L244) use `config/release` and make `config/manager` to use `latest` tag.

stevenhorsman:
> In [#488 (comment)](https://github.com/confidential-containers/operator/pull/488#discussion_r1930071616) I made the comment that the bundle build is a bit confusing. I remember we agreed the approach in this PR but the alternative would be to just make [the bundle creation](https://github.com/confidential-containers/operator/blob/4b6b25f42024efa723b2838bb3f583cd033cbdf2/Makefile#L244) use `config/release` and make `config/manager` to use `latest` tag.

Yep, that would also be completely valid. I think we'd need to update the bundle build process for that, but that's do-able

mythi:
> > In [#488 (comment)](https://github.com/confidential-containers/operator/pull/488#discussion_r1930071616) I made the comment that the bundle build is a bit confusing. I remember we agreed the approach in this PR but the alternative would be to just make [the bundle creation](https://github.com/confidential-containers/operator/blob/4b6b25f42024efa723b2838bb3f583cd033cbdf2/Makefile#L244) use `config/release` and make `config/manager` to use `latest` tag.
> 
> Yep, that would also be completely valid. I think we'd need to update the bundle build process for that, but that's do-able

I'm OK either way. Thoughs @wainersm or others?

wainersm:
Hi @mythi !

> > > In [#488 (comment)](https://github.com/confidential-containers/operator/pull/488#discussion_r1930071616) I made the comment that the bundle build is a bit confusing. I remember we agreed the approach in this PR but the alternative would be to just make [the bundle creation](https://github.com/confidential-containers/operator/blob/4b6b25f42024efa723b2838bb3f583cd033cbdf2/Makefile#L244) use `config/release` and make `config/manager` to use `latest` tag.
> > 
> > 
> > Yep, that would also be completely valid. I think we'd need to update the bundle build process for that, but that's do-able
> 
> I'm OK either way. Thoughs @wainersm or others?

At this point you know more than me and so you are in better position to take a decision, so whatever you think it's better I should sign in.

Nevertheless, I'm trying to understand why we update config/manager on first place. The bundle is built out of `config/manifests` which includes `config/release` as well as `config/samples/ccruntime/default`. It seems wrong to include the later, isn't it? That's why it end up having to update config/manager?

fitzthum (approved):
LGTM

I don't have a strong feeling here but this seems like a straightforward approach.

wainersm (approved):
I suggest to merge this PR as is; @mythi can send an improvement on top of it if he wants to.

mythi:
> Ok, based on the approvals and discussion let's get this in and re-allow the installation of `latest`. If we want to re-work this later, then I'm happy to help.

:+1:

PR #503, opened 2025-02-26
tests/e2e: Fix systemd path handling for Ubuntu 24.04 on s390x

BbolroC:
When upgrading the s390x runner to Ubuntu 24.04, the following error occurred:

```
Destination directory /usr/etc/systemd/system does not exist
```

Previously, the script replaced `/lib` with `/etc` in systemd paths. However, on Ubuntu 24.04, `systemctl show -p FragmentPath docker.service` returns `/usr/lib/systemd/system/docker.service`, rendering the old substitution invalid.
This PR updates the destination directory to
`/etc/systemd/system/docker.service` and fulfills the copy logic if the fragment path starts with either `/lib` or `/usr/lib`.

### DIFF
--- a/tests/e2e/ansible/install_docker.yaml
+++ b/tests/e2e/ansible/install_docker.yaml
@@ -108,8 +108,8 @@
         - name: Copy fragment file to /etc/systemd/system/docker.service.d
           copy:
             src: "{{ fragment_path.stdout }}"
-            dest: "{{ fragment_path.stdout.replace('/lib', '/etc') }}"
-          when: fragment_path.stdout.find('/lib') != -1
+            dest: /etc/systemd/system/docker.service
+          when: fragment_path.stdout.startswith(('/lib', '/usr/lib'))
     - name: Configure `StartLimitBurst=0` in /etc/systemd/system/docker.service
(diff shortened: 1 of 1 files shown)


stevenhorsman:
Do you thinking there is any value in doing something like:
```
ansible_distribution == "Ubuntu" and ansible_distribution_version == "24.04"
```
so that we can support both the 22.04 and 24.04 ways, or do you not see a scenario where we need to go back to 22.04 for s390x e2e tests?

BbolroC:
> Do you thinking there is any value in doing something like:
> 
> ```
> ansible_distribution == "Ubuntu" and ansible_distribution_version == "24.04"
> ```
> 
> so that we can support both the 22.04 and 24.04 ways, or do you not see a scenario where we need to go back to 22.04 for s390x e2e tests?

Yeah, basically going back to an OS <= 22.04 shouldn't happen, but I've considered making the change support any versions.

ldoktor (approved):
Well IIUC this actually shouldn't even break ubuntu2204 as it's hardcoding correct location rather then relying on substitution. Anyway looks good to me.

PR #511, opened 2025-03-05
feat: add CcTolerations field to support custom tolerations in ds

rayanebel:
# Description

👋 Team,

In our infrastructure, we need to configure tolerations for each pod we deploy. Currently, there is no option to add tolerations to the underlying DaemonSets, which prevents us from deploying Confidential Containers.

# Solution

The solution is straightforward: introduce a new field called ccTolerations, allowing us to specify custom tolerations within the DaemonSets.

```yaml
apiVersion: confidentialcontainers.org/v1beta1
kind: CcRuntime
metadata:
  name: ccruntime
spec:
  ccNodeSelector:
    matchLabels:
      node.kubernetes.io/worker: ""
  ccTolerations:
  - effect: NoSchedule
    key: metal
    operator: Exists
  - effect: NoSchedule
    key: protocol
    value: test
    operator: Equal
```

I tested this solution in our cluster, and it works as expected.

```bash
NAME                             DESIRED   CURRENT   READY   UP-TO-DATE   AVAILABLE   NODE SELECTOR                                    AGE
cc-operator-daemon-install       1         1         1       1            1           node.kubernetes.io/worker=""                     36m
cc-operator-daemon-uninstall     0         0         0       0            0           confidentialcontainers.org/startuninstall=true   62m
cc-operator-pre-install-daemon   1         1         1       1            1           node.kubernetes.io/worker=""                     62m
```

```bash
kubectl get ds cc-operator-pre-install-daemon -o yaml | yq '.spec.template.spec.tolerations'

- effect: NoSchedule
  key: metal
  operator: Exists
- effect: NoSchedule
  key: protocol
  operator: Equal
  value: test
  ```

### CODE REVIEW

**Comment by rayanebel** [comment]
@fidencio @bpradipt @wainersm Any chance to review it 🙏

**Review by stevenhorsman — COMMENTED** [review]
It looks reasonable to me, so I've kicked off the tests.

**Comment by AdithyaKrishnan** [comment]
@stevenhorsman @rayanebel
SNP CI was failing initially on this PR but is now passing successfully.
LGTM.

**Review by AdithyaKrishnan — APPROVED** [review]

**Review by mythi — APPROVED** [review]
LGTM

**Review by bpradipt — APPROVED** [review]
Thanks @rayanebel 
It looks good.. 
I think you will also need to re-generate the bundle and commit those

**Comment by rayanebel** [comment]
hey @bpradipt 

When I tried to generate the bundle. I have diff which are not related to my changes. Can you check it ?

**Comment by bpradipt** [comment]
> hey @bpradipt
> 
> When I tried to generate the bundle. I have diff which are not related to my changes. Can you check it ?
Sure. Let me check and revert.

**Comment by stevenhorsman** [comment]
> Unfortunately I'm unable to create the bundle manifest itself with operator-sdk 1.36.1. This has nothing to do with the PR though. @stevenhorsman I see the last bundle update from you. Do you recall which version you used ?

I just ran the instructions in https://github.com/confidential-containers/operator/blob/main/docs/OPERATOR_HUB.md, so it looks like it was `v1.36.1`

**Comment by stevenhorsman** [comment]
So the diffs are just the image version updates?

**Comment by rayanebel** [comment]
@stevenhorsman yes not sure about the result and the changes on `kustomization.yaml`

**Comment by stevenhorsman** [comment]
> @stevenhorsman yes not sure about the result and the changes on `kustomization.yaml`

I think that is fine as it should get reset when we do the next release to that version

**Comment by rayanebel** [comment]
@stevenhorsman @bpradipt is it ok with the latest changes. I can see that one check is failing. Do you know when this will be merged and a new version will be released ?

**Comment by mythi** [comment]
> I can see that one check is failing.

it's a real error cause by this change.

https://github.com/confidential-containers/operator/blob/06863b57207e5a521239e3ff34acb572927ac9f9/.github/workflows/enclave-cc-e2e.yaml#L52

File: config/manager/kustomization.yaml [hunk 1]
```diff
@@ -12,5 +12,5 @@ apiVersion: kustomize.config.k8s.io/v1beta1
 kind: Kustomization
 images:
 - name: controller
-  newName: quay.io/confidential-containers/operator
-  newTag: v0.12.0
+  newName: controller
```
bpradipt, line 15: Should the newName be reverted to `quay.io/confidential-containers/operator` to fix the CI error ?

@mythi @rayanebel ??

AdithyaKrishnan (approved):
LGTM

config/manager/kustomization.yaml
mythi: yes

bpradipt: @rayanebel can you please update the commit to revert `newName` and `newTag`?

rayanebel: @bpradipt done.

bpradipt (approved):
/lgtm

PR #531, opened 2025-07-11
Mirror: add alibaba cloud image mirrors

Xynnn007:
For some users who may encounter the problem of being unable to access `quay.io` when deploying CoCo, we bring an image mirror for coco/kata official image registries.

The sync frequency will be 1~ per day.

Because some company security regulations require that we cannot expose the credentials required to push images to the open source community, we choose to use this method to synchronize images asynchronously. The specific mapping rules are:

quay.io/confidential-containers/operator -> registry-cn-hangzhou.ack.aliyuncs.com/dev/coco-operator
quay.io/kata-containers/kata-deploy  -> registry-cn-hangzhou.ack.aliyuncs.com/dev/coco-kata-deploy
quay.io/confidential-containers/reqs-payload -> registry-cn-hangzhou.ack.aliyuncs.com/dev/coco-reqs-payload
quay.io/kata-containers/kata-deploy-ci   --> registry-cn-hangzhou.ack.aliyuncs.com/dev/coco-kata-deploy-ci

cc @stevenhorsman


BbolroC (approved):
LGTM as far as @stevenhorsman shows an okay sign, thanks @Xynnn007 !

stevenhorsman (approved):
Assuming that the image tags are sync'd in the mirroring this LGTM. In future if when have more regional mirrors we might need a rename/refactor, but this seems good for now.

Do you have any plans to test this mirror upstream, either in our regular automation, or at manually at release time?

Xynnn007:
> Assuming that the image tags are sync'd in the mirroring this LGTM. In future if when have more regional mirrors we might need a rename/refactor, but this seems good for now.
> 
> Do you have any plans to test this mirror upstream, either in our regular automation, or at manually at release time?

i am working on a e2e test for caa on alibaba cloud which depends on this. The nightly test might make sense?

fitzthum (approved):
LGTM. Yeah we may want to have more specifics about the region/mirror name, especially if we have more, but this is fine for now.

Xynnn007:
@mythi Thanks for the solution. I tried two solutions, but they didn't work. 

1. For containerd's mirror, it can only redirect at the registry granularity. In this case, quay.io is redirected to registry-cn-hangzhou.ack.aliyuncs.com. This means that all mirror accesses on quay.io will be redirected to registry-cn-hangzhou.ack.aliyuncs.com, but there will be no mirrors except kata and coco

2. For container registry, this only applies to Podman, Buildah, CRI-O, Skopeo, etc., but not containerd.

I think you are worried about the maintenance cost. I tried to avoid introducing additional maintenance costs for CoCo when writing. You can see that only the address of the image is replaced in kustomization, and the tag and any other configurations are unchanged. This means that even if other parts are changed, as long as the official mirror repository address of kata/coco does not change, the new content will not change.

I can change "mirror" into "alibaba-mirror" or sonething else more specific, to leave the maintaince responsibility to ourselves.

mythi:
> This means that all mirror accesses on quay.io will be redirected to registry-cn-hangzhou.ack.aliyuncs.com, but there will be no mirrors except kata and coco

Are there images that *can* be pulled from `quay.io` and this mirror config would break it? 

Would `ghcr.io` work better for you? Perhaps we can move the operator images there. Kata seems to be available already.

Xynnn007:
@mythi Actually, on alibaba cloud, if we want to connect to quay.io, we need to have a VPN. Some users will use images on quay.io but usually not. For users who only want to use coco/kata image on quay.io this is not the case.

ghcr.io is the same as quay.io.

mythi:
It is possible to have downstream kustomizations in a separate repo and let them pull from this repo:

```
apiVersion: kustomize.config.k8s.io/v1beta1
kind: Kustomization
resources:
- "github.com/confidential-containers/operator/config/release?ref=v0.14.0"

images:
- name: quay.io/confidential-containers/operator
  newName: registry-cn-hangzhou.ack.aliyuncs.com/dev/coco-operator
```

and then something like `kubectl apply -k github.com/alibaba/coco-cca-install` but it pulls the base from here.

Xynnn007:
@mythi thanks. Let me take a try.

PR #532, opened 2025-07-22
Release: Update nydus snapshotter version

niteeshkd:
This is needed for the step#4 of [the checklist v0.15.0](https://github.com/confidential-containers/confidential-containers/issues/311) . 

The snapshotter version used by the operator needs to match with the [snapshotter used by kata](https://github.com/kata-containers/kata-containers/blob/main/versions.yaml#L350).

PR #539, opened 2025-09-24
Release: Update payloads for v0.16.0

niteeshkd:
PR for step#5 of [checklist for v0.16.0](https://github.com/confidential-containers/confidential-containers/pull/318)

### DIFF
--- a/.github/workflows/ccruntime_e2e.yaml
+++ b/.github/workflows/ccruntime_e2e.yaml
@@ -44,6 +44,29 @@ jobs:
         env:
           TARGET_BRANCH: ${{ inputs.target-branch }}
 
+      - name: Remove unnecessary directories to free up space
+        run: |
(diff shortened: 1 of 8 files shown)


fitzthum:
Ubuntu 22.04 test is failing. Looks like we're low on disk space. Not sure what the root cause is. 

cc @wainersm

mythi:
`NYDUS_SNAPSHOTTER_VERSION = v0.15.4`?

mythi:
> > `NYDUS_SNAPSHOTTER_VERSION = v0.15.4`?
> 
> Are you suggesting to bump the nydus snapshotter version? It'd require a bump on the pre-req daemonset, and I don't think that scenario (the update) was tested before?

The version bump is on our release checklist. We've forgotten do it a few times so I take my comment as a reminder. @niteeshkd did something similar in #532.

fitzthum (approved):
LGTM

let's keep a close eye on the post-merge jobs

